PR #3, opened 2019-02-19
Add clean command and pretty error output

jonpas:
- Add `hemtt clean [--force]` command:
  - Removes all PBOs (and other generated files, but there are none of those now)
  - `--force` removes all releases as well
- Remove release of same version on `hemtt build --release --force` to assure release build is clean (show warning and fail otherwise)
- Alphabetized `mod` list
- Add pretty error output and macro (same as in armake2~, this project has no license yet so have to ask @KoffeinFlummi about it~)
  - Can be removed, went out of the scope of this PR a bit, but I believe it would be neat to display errors/warnings in a similar style, or at least to avoid panic where errors are gracefully handled
  - Windows ANSI support added to colored (lacking it)


jonpas:
Colored might not work on Windows CMD, need to test.

jonpas:
Explicit permission from KoffeinFlummi acquired, "proof":
```
jonpas [05:53]
 Q: Do I have permission to use code from `error.rs` in that HEMTT project? it's currently unlicensed
 and what kind of attribution do you want if yes?

koffeinflummi [07:03]
 of course, and don't care.
```

src/build.rs
```diff
@@ -37,41 +38,30 @@ pub fn build(p: &crate::project::Project) -> Result<(), std::io::Error> {
...
-pub fn release(p: &crate::project::Project) -> Result<(), std::io::Error> {
-    let mut stdout = StandardStream::stdout(ColorChoice::Always);
-    let version = match &p.version {
-        Some(v) => v,
-        None => panic!("Unable to determine version number"),
-    };
-    println!("Building Release Version: {}", version);
+pub fn release(p: &crate::project::Project, version: &String) -> Result<(), Error> {
+    println!(" {} release v{}", "Preparing".green().bold(), version);
     build(&p)?;
```
synixebrett: Might as well move the build down below the error, otherwise they will need to build it twice.

synixebrett:
We will continue using the PR while it is WIP. Once it is ready you will need to create a new one going into `master`. We will then just post the PR in the other one to save the comments.

jonpas:
@synixebrett done. Also changed base of the PR to `master` (GitHub can do that), no problems since it has same history.

synixebrett:
Ah cool

PR #12, opened 2019-02-19
Check include folders and expose to configuration

jonpas:
- Fix building when `./include` folder does not exist (eg. no CBA includes or different name)
- Expose include folders to `hemtt.json`
- Print pretty errors from armake2 and gracefully handle them
- Move `keys` from root to `releases/keys`
  - Shall be made configurable one day

### DIFF
--- a/docs/json.md
+++ b/docs/json.md
@@ -84,6 +84,19 @@ HEMTT will copy the files to the release directory after a successful release bu
 ```
 <hr/>
 
+## include
+**Type**: Array \[String (Path)\]
+
+HEMTT will include matching relative or absolute paths when building.
+
+```json
+"include": [
+    "./include"
+]
+```
+
+`./include` will be automatically added on project creation if "include" folder is present.
+
 ## exclude
 **Type**: Array \[String\]
 
--- a/src/build.rs
+++ b/src/build.rs
@@ -10,6 +10,7 @@ use std::path::{Path, PathBuf};
 use std::time::{Duration, SystemTime};
 
 use crate::error;
+use crate::error::*;
 
 pub fn modtime(addon: String) -> Result<SystemTime, std::io::Error> {
     let mut recent: SystemTime = SystemTime::now() - Duration::new(60 * 60 * 24 * 365 * 10, 0);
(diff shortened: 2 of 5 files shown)


src/build.rs
```diff
@@ -71,15 +78,16 @@ pub fn release(p: &crate::project::Project, version: &String) -> Result<(), Erro
...
-        fs::create_dir("keys")?;
+    if !Path::new("releases/keys").exists() {
+        fs::create_dir("releases/keys")?;
     }
-    if !Path::new(&format!("keys/{}.bikey", p.prefix)).exists() {
+    if !Path::new(&format!("releases/keys/{}.bikey", p.prefix)).exists() {
+        println!(" {} {}.bikey", "Generating".green().bold(), p.prefix);
         armake2::sign::cmd_keygen(PathBuf::from(&p.prefix))?;
-        fs::rename(format!("{}.bikey", p.prefix), format!("keys/{}.bikey", p.prefix))?;
-        fs::rename(format!("{}.biprivatekey", p.prefix), format!("keys/{}.biprivatekey", p.prefix))?;
```
synixebrett: Should the private key be moved into releases? I thought they could be used for multiple releases. This way you could keep the same public key on servers so you know an update came from the same source as your last version.

jonpas: I think it's better to have them in releases folder (next to individual, yeah), for easy cleanup and isolated from other things.

synixebrett: Yeah this is fine, I was just confused for a minute.

docs/json.md
```diff
@@ -84,6 +84,19 @@ HEMTT will copy the files to the release directory after a successful release bu
...
+
+HEMTT will include matching relative or absolute paths when building.
+
+```json
+"include": [
+    "./include"
+]
+```
+
+`./include` will be automatically added on project creation if "include" folder is present.
```
jonpas: We can remove that functionality if wanted though, or change that one to be checked in runtime, but I don't like that.

jonpas:
Documentation added.

docs/json.md
synixebrett: Maybe use `-include` inside the hemtt.json file to specify not including it?

jonpas: It will only be added if the actual folder is present during `hemtt create`, and actually added to JSON itself. So if projects removes it they just remove it from the JSON.

synixebrett: Ah ok, didn't notice on project creation

PR #23, opened 2019-02-20
Optional Addons

### LINKED ISSUE #18: Build Feature - Optional addons/components
Allow building addons in optionals folder.

Always build during release and copy accordingly.

Build with `build` when specified, eg. `build --opt <optional>` or `build --opts` for all.

synixebrett:
- Add `--opts` to `build`
- Add `optionals` to `hemtt.json`
- Close #18

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -381,7 +381,7 @@ dependencies = [
 
 [[package]]
 name = "hemtt"
-version = "0.5.0"
+version = "0.4.1"
 dependencies = [
  "ansi_term 0.11.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "armake2 0.3.0 (git+https://github.com/KoffeinFlummi/armake2)",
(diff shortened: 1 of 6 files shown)


jonpas:
Are optional getting built in the optional folder (PBOs put there)?

I was thinking --opts would copy them to /addons as well, for quick testing procedures, but for that it has to not build them unless specified, and also clean them up from /addons when not specified.

synixebrett:
Hm, I hadn't considered moving them into addons. I'm not sure how we should do this. You'd likely want different behaviours for `build` and `build --release` and right now the build process is the same code for both.

jonpas:
Not really, release still builds them, just copies them to a different place in releases/version/optionals.

Build on the other hand builds them in optionals and copies to addons when requested.

I am not sure how to do this in an ideal way yet, will think about it further and write in the new issue.

Issue #17, opened 2019-02-19
ACE3 Configurability


jonpas:
Some required configuration to satisfy CBA/ACE3/ACRE2 and other projects I want to use this project on:
- [x] Header extensions (prefix and other user-made ones) - example: `-e "version=<git-hash>"` - read from `hemtt.json` and straight passed to armake2's `cmd_build` -> #30
- [x] Signature name (`-s <signature>` in armake2) - just use instead of middle part -> #30
- [x] `@mod` folder name - template setup -> #30
- [x] Key name (bikey) - template setup -> #30
- [x] Release archive name (ACE3 does `ace3_3.1.0.zip`, CBA does `CBA_A3_v3.1.0.zip`) to zip utility - template setup -> #9 
- [x] Optionals release build path (launchable  optional mod folders) -> #132, #225 

Other:
- Disable specific warnings (`-w --warning <wname>` in armake2) - add to armake2's list of ignored warnings -> #197
- ~~Build folder (eg. armake setup with ACE3 builds into `@ace/addons` and also copies other files, that way you get a closer to release build preview, releases are probably best kept completely separate, or also not, HEMTT could resolve that easily)~~

jonpas:
#197 was moved out of this issue.

I am considering this done, all ACE3 requirements are now satisfied.

PR #30, opened 2019-02-22
Add configurable header extensions, signature name and mod name

jonpas:
- Work on #17 
- Add configurable header extensions
- Add configurable signature name (the middle part between PBO name and ".bisign")
- Add configurable mod name (`@mod` folder name)
- Add configurable key name (`.bikey` and `.biprivatekey` names)
- Add initial replacement patterns for header extensions and signature name properties (will be expanded in the future either with custom handling or [handlebars](https://github.com/sunng87/handlebars-rust) - thus the syntax)
- Change signing to print pretty errors from armake2 instead of panic on failure

Git hash parser can be added in separate PR as utility, otherwise this will get too big. Zip utility will also need it for parsing zip name.

### DIFF
--- a/docs/json.md
+++ b/docs/json.md
@@ -154,3 +154,13 @@ HEMTT will use the specified signature name as part of the full signature (`.bis
 ### Replacements
 
 - `{{version}}`: Mod version in format `MAJOR.MINOR.PATCH.BUILD`
+
+### Example
+
+```json
+"project": "TST",
+"version": "1.0.0.0",
+"signame": "my-{{version}}"
+```
+
+Above will result in signature name of `TST_<addon>.pbo.my-1.0.0.0.bisign` (where `<addon>` is the name of the addon folder), located next to the matching addon PBO.
--- a/docs/json.md
+++ b/docs/json.md
@@ -154,6 +154,28 @@ HEMTT will use the specified mod name (without `@`) to form `@mod` folder.
 "modname": "my_mod"
 ```
 
+## keyname
+**Type**: String
+
+HEMTT will use the specified key name for `.bikey` and `.biprivatekey` names.
+
+```json
+"keyname": "my_key"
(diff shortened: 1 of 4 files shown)


docs/json.md
```diff
@@ -121,3 +121,36 @@ HEMMT will build the specified optionals from the `./optionals` folder.
...
+
+## signame
+**Type**: String
+
+HEMTT will use the specified signature name as part of the full signature (`.bisign`) name.
+
+```json
+"signame": "my_custom_name"
+```
+
```
synixebrett: It would be nice to put a file tree here showing the final name and place of the file for people that are less familiar with signing.

docs/json.md
```diff
@@ -121,3 +121,36 @@ HEMMT will build the specified optionals from the `./optionals` folder.
...
 ```
+
+## headerexts
+**Type**: Array \[String\]
+
+HEMTT will apply specified header extensions to each PBO.
+
+```json
+"headerexts": [
+    "author=me"
```
synixebrett: We should probably be doing author ourselves since we already store the author in the project file.

jonpas: No, this is a bad example maybe, but we shouldn't store anything ourselves in there, all it adds is PBO size. We can support author with a replacement pattern though. 

ACE/CBA store git hash here for issue reporting purposes (gets displayed in RPT), that's what I will use it for primarily, but git hash part has yet to be done.

jonpas: Btw, those docs will get updated with possible replacement patterns when I add git hash utility. Example will be better then!

jonpas:
Fixed conflicts, also added key name.

Zip name is still missing, but I will add that in the other PR with zip utility, which will actually call it, so we don't have unused functions warnings in the meantime.

I consider this PR done for now.

Issue #42, opened 2019-02-25
Don't write private key to disk


bovine3dom:
Related to #34 - I can't think of any good reason why the private key should be written to disk, only downsides.

koffeinflummi suggests just leaving it in memory and bypassing `cmd_sign`:

```rust
let key = BIPrivateKey::generate(1024, "myaddon".to_string);
let public_key = key.to_public_key();
public_key.write(File::create("whatever.bikey"));

let sig = key.sign(my_pbo)
sig.write(File::create("whatever.bisign")
```

This would be a nice little unique feature for HEMTT :)

PR #47, opened 2019-02-27
Build Refactor

### LINKED ISSUE #36: Can't skip specified addons
`./hemtt build spectator,zeus --skip spectator`
Spectator will still be built.

While something like this shouldn't really ever be ran, we should at least print a warning. Interested in @jonpas's opinion on if it should still be skipped or not.

### LINKED ISSUE #39: Optionals do not get cleaned
Bug since building optionals was added, they don't get cleaned with `hemtt clean` nor when rebuilding with `--force`.

synixebrett:
- Close #39 
- Close #36 
- Add a progress bar during PBO buildin
- Allow `hemtt.toml`, uses toml if both toml and json exist
- Change build output:


jonpas:
signing.rs to sign.rs maybe?

I like it, would like to test before merge though.

Side-note: we should default to building all optionals when argument not passed.

synixebrett:
I think just using `optionals` inside `hemtt.json` will be fine, but for release builds we could default to `all`. Since during development you may not want to wait for optionals you are not testing.

bovine3dom:
I think "pub key" is clearer than KeyGen. Also, perhaps not the fault of this PR, shouldn't the public key have a version number appended to it?

synixebrett:
Signing is going to need its own refactor I think, especially with #42

jonpas:
Key can have a version number with my configurability PR. Maybe we should make it default though especially since it's regenerated every release.

synixebrett:
I'm working on templating right now, we could have a default of `{prefix}.{version}`

jonpas:
I prefer underscore, which is also in-line with current CBA/ACE.

src/project.rs
```diff
@@ -48,8 +49,13 @@ fn default_include() -> Vec<PathBuf> {
 
 impl Project {
     pub fn save(&self) -> Result<(), Error> {
-        let mut out = File::create(crate::HEMTT_FILE)?;
-        out.write_fmt(format_args!("{}", serde_json::to_string_pretty(&self)?))?;
+        let file = path().unwrap();
```
jonpas: This panics instead of printing pretty error from `path()`.
```suggestion
        let file = path().print_error(true);
```

synixebrett:
I'm thinking of keeping JSON default just because there is a higher chance someone has seen it before. They won't need to worry about TOML to use HEMTT. If someone wants to use TOML instead of JSON it is really quick to replace the default `hemtt.json` file.

HEMTT will use TOML over JSON if both files are present.

jonpas:
Sounds good.

Sorry, correct code that works for printing good error on that is:
```rust
        let file = path().print_error(true).unwrap();
```

And I meant `hemtt init` above, not `create` (edited now), doesn't work as it should with this PR.

jonpas:
Subjective, but I prefer this progress bar style, I think it's a lot more readable and simpler:
```
[#######################################                           ] [ 59.00%] [00:06]
```

jonpas:
This style fits even more (directly from cargo):
```
Building [==================================================>    ] 200/224: hemtt(bin) 
```

Would look like this for HEMTT
```
Building [==================================================>    ] 200/224: ace(release)
Building [==================================================>    ] 200/224: ace(build)
```

src/main.rs
```diff
@@ -147,81 +153,152 @@ fn run_command(args: &Args) -> Result<(), Error> {
...
+        state.stage = crate::state::Stage::PostBuild;
+        state.result = Some(&result);
+        p.run(&state).unwrap_or_print();
+        if args.flag_release {
+            build::release::release(&p, &version).unwrap_or_print();
+            state.stage = crate::state::Stage::ReleaseBuild;
+            p.run(&state).unwrap_or_print();
+            println!("  {} {} v{}", match result.failed.len() {
+                0 => "Finished".green().bold(),
+                _ => "Finished".yellow().bold(),
```
jonpas: Should this be `error: Could not finish` (`error:` in red) like Cargo on compile error?

synixebrett: I avoided red because it doesn't stop on failure. If we changed HEMTT to stop on failure red would be used, but I didn't feel a big `error` in red was suitable for a partial fail. That actual PBO that failed gets a red `error:`.

jonpas: True, yellow is fine. Maybe add a note that not all PBOs were successful?

PR #48, opened 2019-02-27
Templating

synixebrett:
- Add basic templating for `hemtt.json` using [handlebars](https://github.com/sunng87/handlebars-rust)
- Replace `print_error(true)` with `unwrap_or_print()`
- Replace `print_error(false)` with `print()`

### DIFF
--- a/src/project.rs
+++ b/src/project.rs
@@ -1,12 +1,16 @@
+use chrono::prelude::*;
 use serde::{Serialize, Deserialize};
 use serde_json;
 
+use std::collections::BTreeMap;
 use std::fs::File;
 use std::io::{Write, Error};
 use std::io::BufReader;
 use std::io::prelude::*;
 use std::io::Read;
 use std::path::{Path, PathBuf};
+
+use crate::template::render;
 
 #[derive(Serialize, Deserialize)]
 pub struct Project {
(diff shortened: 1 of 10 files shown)


synixebrett:
Date helper is using [chrono specifiers](https://docs.rs/chrono/0.4.6/chrono/format/strftime/index.html#specifiers) for formatting

```json
"signame": "{{version}}-{{date \"%y%m%d\"}}"
```
That would result in something like `ACE_zeus.pbo.3.12.5.40-190227.bisign`

src/project.rs
```diff
@@ -143,10 +148,15 @@ pub fn get_project() -> Result<Project, Error> {
...
-    let p: Project = match toml_exists() {
+    let mut p: Project = match toml_exists() {
         true => toml::from_str(contents.as_str()).unwrap(),
         false => serde_json::from_str(contents.as_str())?
     };
+    p.template_data = BTreeMap::new();
+    p.template_data.insert("name", p.name.clone());
+    p.template_data.insert("prefix", p.prefix.clone());
+    p.template_data.insert("author", p.author.clone());
+    p.template_data.insert("version", p.version.clone().unwrap());
```
jonpas: Would be nice if we could define those in project file as well, use some project-specific tools for example.
```json
"templates": [
    "my_id": "exec get_my_id.sh"
]
```

This could use same system of execution as custom build steps, which we still have to figure out (especially to support both Windows and Linux we need some kind of separation).

synixebrett: This is planned, I'm waiting for Build Steps so I can keep the Windows / Linux separation consistent between templating and build steps, since build steps is going to be more complicated I want to tackle that first.

jonpas: Yup, my thinking as well. :+1:

Issue #38, opened 2019-02-22
`rustfmt` as a pre-commit hook?


bovine3dom:
E.g, https://github.com/google/tarpc/blob/master/hooks/pre-commit

We do something similar in another project I work on. It works rather well for ensuring that everyone sticks to the same code style.

jonpas:
rustfmt is sometimes really weird, but if that's the Rust standard = :+1:.

jonpas:
Closed by #77:
> I can't see any nice way of distributing pre-commit hooks, so I gave up on that idea.

PR #51, opened 2019-03-02
Don't write private key to disk

### LINKED ISSUE #42: Don't write private key to disk

bovine3dom:
Closes #42.

~~I haven't left writing the private key to disk as an option. The only valid usage I can think of is if you needed to distribute signing across a server farm, which you don't.~~

~~I've cheekily made the public key follow the `name_version.bikey` format since there will be a different one for each version now.~~

~~I've just spotted #47. I'm happy to rebase this on that once it has been merged.~~

---

- Don't write private key to disk unless `reuse_private_key` is set to true in the project configuration
    - If it is set to true, revert to the old behaviour
- Change the default key name to include the version if `reuse_private_key` is false, stick with old default otherwise

- Split copy / sign up into separate functions and do a bit of tidying up (I went a bit off-piste, sorry...)

### DIFF
--- a/docs/project.md
+++ b/docs/project.md
@@ -168,6 +168,8 @@ HEMTT will use the specified mod name (without `@`) to form `@mod` folder. Suppo
 
 HEMTT will use the specified key name for `.bikey` and `.biprivatekey` names. Supports [templating](/templating.md).
 
+The default is `{{prefix}}_{{version}}`, unless `reuse_private_key` is `true`, in which case the default is set to `{{prefix}}`
+
 ```json
 "keyname": "my_key"
 ```
@@ -177,10 +179,11 @@ HEMTT will use the specified key name for `.bikey` and `.biprivatekey` names. Su
 ```json
 "project": "TST",
 "version": "1.0.0.0",
-"keyname": "my_key-{{version}}"
+"keyname": "my_key_{{version}}"
 ```
 
-Above will result in key name of `my_key-1.0.0.0.bikey` and private key name of `my_key-1.0.0.0.biprivatekey`.
+Above will result in key name of `my_key_1.0.0.0.bikey` and private key name of `my_key_1.0.0.0.biprivatekey`.
+
 
 ## signame
 **Type**: String
@@ -200,3 +203,16 @@ HEMTT will use the specified signature name as part of the full signature (`.bis
 ```
 
 Above will result in signature name of `TST_<addon>.pbo.my-1.0.0.0.bisign` (where `<addon>` is the name of the addon folder), located next to the matching addon PBO.
+
+
+## reuse_private_key
+
+**Type**: bool
+
+If set to `true`, HEMTT will use (and re-use) `releases/keys/{keyname}.biprivatekey`. It will be generated if it doesn't exist.
+
+The default behaviour is to generate a new private key each time and discard it immediately; HEMTT strongly recommends that you only re-use the key if you are making a client-side mod where it will not matter if clients are running different versions of the mod.
+
+```json
+"reuse_private_key": false
+```
--- a/docs/project.md
+++ b/docs/project.md
@@ -168,7 +168,12 @@ HEMTT will use the specified mod name (without `@`) to form `@mod` folder. Suppo
 
 HEMTT will use the specified key name for `.bikey` and `.biprivatekey` names. Supports [templating](/templating.md).
 
-The default is `{{prefix}}_{{version}}`, unless `reuse_private_key` is `true`, in which case the default is set to `{{prefix}}`
+The default is set according to the following table:
+
+| `reuse_private_key` value | Default `keyname`         |
+| ------------------------- | ------------------------- |
+| `false`                   | `{{prefix}}_{{version}}`  |
+| `true`                    | `{{prefix}}`              |
 
 ```json
 "keyname": "my_key"
@@ -211,7 +216,9 @@ Above will result in signature name of `TST_<addon>.pbo.my-1.0.0.0.bisign` (wher
 
 If set to `true`, HEMTT will use (and re-use) `releases/keys/{keyname}.biprivatekey`. It will be generated if it doesn't exist.
 
-The default behaviour is to generate a new private key each time and discard it immediately; HEMTT strongly recommends that you only re-use the key if you are making a client-side mod where it will not matter if clients are running different versions of the mod.
+The default behaviour is to generate a new private key each time and discard it immediately.
+
+!> HEMTT strongly recommends that you only re-use the key if you are making a client-side mod where it will not matter if clients are running different versions of the mod.
 
 ```json
 "reuse_private_key": false
--- a/src/build/release.rs
+++ b/src/build/release.rs
@@ -9,8 +9,8 @@ use std::io::Error;
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, Mutex};
 
-use crate::error::*;
 use crate::build::sign;
+use crate::error::*;
 
 pub fn release(p: &crate::project::Project, version: &String) -> Result<(), Error> {
     let modname = p.get_modname();
@@ -37,17 +37,13 @@ pub fn release(p: &crate::project::Project, version: &String) -> Result<(), Erro
     let keyname = p.get_keyname();
     // Generate and store key if required
     if p.reuse_private_key {
-
         // Make a new keypair if there isn't one already
         if !Path::new(&format!("releases/keys/{}.bikey", keyname)).exists() {
             println!("    {} {}.bikey", "KeyGen".green().bold(), keyname);
 
             // Generate and write the keypair to disk in the current directory
             cmd_keygen(PathBuf::from(&keyname))?;
-            fs::rename(
-                format!("{}.bikey", keyname),
-                format!("releases/keys/{}.bikey", keyname),
-            )?;
+            fs::rename(format!("{}.bikey", keyname), format!("releases/keys/{}.bikey", keyname))?;
             fs::rename(
                 format!("{}.biprivatekey", keyname),
                 format!("releases/keys/{}.biprivatekey", keyname),
@@ -56,8 +52,7 @@ pub fn release(p: &crate::project::Project, version: &String) -> Result<(), Erro
 
         // Read the private key from disk
         key = BIPrivateKey::read(
-            &mut File::open(format!("releases/keys/{}.biprivatekey", keyname))
-                .expect("Failed to open private key"),
+            &mut File::open(format!("releases/keys/{}.biprivatekey", keyname)).expect("Failed to open private key"),
         )
         .expect("Failed to read private key");
     } else {
@@ -66,9 +61,7 @@ pub fn release(p: &crate::project::Project, version: &String) -> Result<(), Erro
         let public_key = key.to_public_key();
 
         // Write the public key to disk
-        public_key.write(
-            &mut std::fs::File::create(format!("releases/keys/{}.bikey", keyname)).unwrap_or_print(),
-        )?;
+        public_key.write(&mut std::fs::File::create(format!("releases/keys/{}.bikey", keyname)).unwrap_or_print())?;
     }
 
     // Copy public key to specific release dir
@@ -94,9 +87,7 @@ pub fn release(p: &crate::project::Project, version: &String) -> Result<(), Erro
 
     folder = String::from("optionals");
     if Path::new(&folder).exists() {
-        let addonsfolder = iformat!(
-            "releases/{version}/@{modname}/{folder}", version, modname, folder
-        );
+        let addonsfolder = iformat!("releases/{version}/@{modname}/{folder}", version, modname, folder);
         if !Path::new(&addonsfolder).exists() {
             fs::create_dir_all(addonsfolder)?;
         }
@@ -106,7 +97,7 @@ pub fn release(p: &crate::project::Project, version: &String) -> Result<(), Erro
             .collect();
         opts.par_iter().for_each(|entry| {
             // TODO split copy and sign
-            // for copying, we need to know source path, addons folder and pbo_filename 
+            // for copying, we need to know source path, addons folder and pbo_filename
             // (we could get this but that seems like extra faff)
             // for signing, we need to know addons folder, PBO file name and key
             if sign::copy_sign(&folder, &entry.path(), &p, &key).unwrap_or_print() {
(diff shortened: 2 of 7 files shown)


synixebrett:
Looks good, I should have the build refactor done by the end of this next week so I am going to wait until after that to merge this.

src/build/release.rs
```diff
@@ -22,26 +22,24 @@ pub fn release(p: &crate::project::Project, version: &String) -> Result<(), Erro
...
-    let keyname = p.get_keyname();
-    if !Path::new(&format!("releases/keys/{}.bikey", keyname)).exists() {
-        println!("    {} {}.bikey", "KeyGen".green().bold(), keyname);
-        armake2::sign::cmd_keygen(PathBuf::from(&keyname))?;
-        fs::rename(format!("{}.bikey", keyname), format!("releases/keys/{}.bikey", keyname))?;
-        fs::rename(format!("{}.biprivatekey", keyname), format!("releases/keys/{}.biprivatekey", keyname))?;
-    }
 
-    // Sign
+    let keyname = format!("{}.{}",p.get_keyname(),version);
```
jonpas:
```suggestion
    // KeyGen
    let keyname = format!("{}.{}",p.get_keyname(),version);
```

src/build/sign.rs
```diff
@@ -1,30 +1,34 @@
...
     }
 
     let modname = p.get_modname();
-    fs::copy(&cpath, format!("releases/{}/@{}/{}/{}", version, modname, folder, pbo))?;
+    fs::copy(&cpath, format!("releases/{}/@{}/{}/{}", version, modname, folder, pbo_path))?;
+
+    let signame = p.get_signame(&pbo_path);
+
+    let pbo = armake2::pbo::PBO::read(&mut 
+                                      std::fs::File::open(PathBuf::from(
```
jonpas: What the hell is this indentation! Use variables. :laughing:

bovine3dom: #38 :(

I'm not used to thinking about indentation. I think Vim probably did it for me. I don't really want to make loads of variables when all it's really doing is updating the PBO path. Rust makes it really verbose to do that 🤷‍♂️

bovine3dom:
Rebased on current master, should be ready for merge. 

I was getting some very strange errors when I ran it but I think that was because I was changing git branches while using `entr` to rebuild the files as I changed them... I haven't been able to reproduce the errors since, so I don't think they were real. I've since run it quite a few times and used `armake2 verify` to check a few. All seems fine.

src/build/sign.rs
jonpas: It's a lot easier to change variables and move them around than 3x nested function call.

src/build/sign.rs
```diff
@@ -1,30 +1,48 @@
...
+    let pbo_path = cpath.replace((folder.clone() + "/").as_str(), "");
+    if !path.ends_with(".pbo") && !pbo_path.contains(p.prefix.as_str()) {
         return Ok(false);
     }
 
     let modname = p.get_modname();
-    fs::copy(&cpath, format!("releases/{}/@{}/{}/{}", version, modname, folder, pbo))?;
+    fs::copy(
+        &cpath,
+        format!("releases/{}/@{}/{}/{}", version, modname, folder, pbo_path),
```
jonpas: :thinking:

bovine3dom:
I'll add a `reuse_private_key` option to the TOML and maybe a flag.

src/build/sign.rs
bovine3dom: What's the `thunk` for? `rustfmt` decided on the newline, not me :(

jonpas: It's hideous. :D

src/build/sign.rs
```diff
@@ -11,36 +11,39 @@ pub fn copy_sign(
...
 ) -> Result<bool, Error> {
     let path = entry.path();
-    let cpath = path.clone();
-    let cpath = cpath.to_str().unwrap().replace(r#"\"#, "/");
-    let pbo_path = cpath.replace((folder.clone() + "/").as_str(), "");
-    if !path.ends_with(".pbo") && !pbo_path.contains(p.prefix.as_str()) {
+    let path_posix = path.clone().to_str().unwrap().replace(r#"\"#, "/");
+    let pbo_filename = path_posix.replace((folder.clone() + "/").as_str(), "");
+
+    if !path.ends_with(".pbo") && !pbo_filename.starts_with(p.prefix.as_str()) {
```
bovine3dom: This is where I integrated #51. I meant to separate it into a separate commit but apparently I forgot.

bovine3dom:
I think this is ready to be reviewed.

Is it worth documenting the weird behaviour you'll get if you specify a dynamic `keyname` but ask us to reuse the key?

synixebrett:
Yeah, we should probably use a static name then for the private key. The only reason it would be on disk is if they wanted to reuse it. In that case they would want a static name.

docs/project.md
```diff
@@ -167,6 +167,8 @@ HEMTT will use the specified mod name (without `@`) to form `@mod` folder. Suppo
 
 HEMTT will use the specified key name for `.bikey` and `.biprivatekey` names. Supports [templating](/templating.md).
 
+The default is `{{prefix}}_{{version}}`, unless `reuse_private_key` is `true`, in which case the default is set to `{{prefix}}`
```
synixebrett: It would be nice if this had a table instead, showing the difference between `reuse_private_key` being true or false. This would help clearly show what the default is, and what the output will be in a clear and readable way.

docs/project.md
```diff
@@ -199,3 +202,16 @@ HEMTT will use the specified signature name as part of the full signature (`.bis
...
 Above will result in signature name of `TST_<addon>.pbo.my-1.0.0.0.bisign` (where `<addon>` is the name of the addon folder), located next to the matching addon PBO.
+
+
+## reuse_private_key
+
+**Type**: bool
+
+If set to `true`, HEMTT will use (and re-use) `releases/keys/{keyname}.biprivatekey`. It will be generated if it doesn't exist.
+
+The default behaviour is to generate a new private key each time and discard it immediately; HEMTT strongly recommends that you only re-use the key if you are making a client-side mod where it will not matter if clients are running different versions of the mod.
```
synixebrett: docisfy has a nice helper for highlighting important info. https://docsify.js.org/#/helpers?id=important-content

It should be used here to highlight the recommendation.

src/build/release.rs
```diff
@@ -18,45 +22,87 @@ pub fn release(p: &crate::project::Project, version: &String) -> Result<(), Erro
...
+                format!("releases/keys/{}.bikey", keyname),
+            )?;
+            fs::rename(
+                format!("{}.biprivatekey", keyname),
+                format!("releases/keys/{}.biprivatekey", keyname),
+            )?;
+        }
+
+        // Read the private key from disk
+        key = BIPrivateKey::read(
```
synixebrett: I am not a fan of this formatting at all, is this what rustfmt did?

src/build/release.rs
```diff
@@ -18,45 +22,87 @@ pub fn release(p: &crate::project::Project, version: &String) -> Result<(), Erro
...
             *count.lock().unwrap() += 1;
         }
     });
 
     folder = String::from("optionals");
     if Path::new(&folder).exists() {
-        if !Path::new(&format!("releases/{}/@{}/{}", version, modname, folder)).exists() {
-            fs::create_dir_all(format!("releases/{}/@{}/{}", version, modname, folder))?;
+        let addonsfolder = format!(
+            "releases/{ver}/@{mod}/{addons}", ver=version, mod=modname, addons=folder
```
synixebrett: The use of variables here seems unnecessary, especially since we don't use them anywhere else. `mod` is also a keyword and shouldn't be used as a variable.

src/build/release.rs
```diff
@@ -18,45 +22,87 @@ pub fn release(p: &crate::project::Project, version: &String) -> Result<(), Erro
...
         }
-        let opts: Vec<_> = fs::read_dir(&folder).unwrap()
+        let opts: Vec<_> = fs::read_dir(&folder)
+            .unwrap()
             .map(|file| file.unwrap())
             .collect();
         opts.par_iter().for_each(|entry| {
             // TODO split copy and sign
-            if crate::build::sign::copy_sign(&folder, &entry, &p, &version).unwrap() {
+            // for copying, we need to know source path, addons folder and pbo_filename (we could
```
synixebrett: move the start of ( to the next line

src/build/sign.rs
```diff
@@ -1,30 +1,59 @@
...
+pub fn copy(
+    path_posix: &String,
+    addonsfolder: &String,
+    pbo_filename: &String,
+) -> Result<bool, Error> {
+    fs::copy(path_posix, format!("{}/{}", addonsfolder, pbo_filename))?;
+    Ok(true)
+}
+
+pub fn sign(
```
synixebrett: I know this is rustfmt's doing, and the rusty way.

Personally, I am not a fan, but I would be open to switching the entire project to it. That is out of scope here though.

src/build/sign.rs
```diff
@@ -1,30 +1,59 @@
...
         return Ok(false);
     }
 
     let modname = p.get_modname();
-    fs::copy(&cpath, format!("releases/{}/@{}/{}/{}", version, modname, folder, pbo))?;
 
-    let signame = p.get_signame(&pbo);
-    let keyname = p.get_keyname();
+    let addonsfolder = format!(
+        "releases/{ver}/@{mod}/{addons}", ver=p.version.clone().unwrap(), mod=modname, addons=folder
```
synixebrett: Same as previous variables comment

src/project.rs
```diff
@@ -63,6 +63,8 @@ pub struct Project {
 
     #[serde(skip_deserializing,skip_serializing)]
     pub template_data: BTreeMap<&'static str, String>,
+
+    pub reuse_private_key: Option<bool>,
```
synixebrett: This should be above the internal variable `template_data`

synixebrett: This should use a serde default flag instead of an `Option` since it will always have a value.

src/project.rs
```diff
@@ -93,7 +95,11 @@ impl Project {
 
     pub fn get_keyname(&self) -> String {
         if self.keyname.is_empty() {
-            self.prefix.clone()
+            if self.reuse_private_key.unwrap_or(false) {
+                self.prefix.clone()
+            } else {
+                format!("{}_{}", &self.prefix, &self.version.clone().unwrap())
```
synixebrett: `unwrap` is added a few times in this PR, make sure `unwrap_or_print` is used wherever allowed. That is what prints the nice readable error messages rather.

src/build/release.rs
bovine3dom: Yeah, I wouldn't have done this. It's before I increased the line length in the other PR so it'll probably disappear if I run it again.

src/build/release.rs
bovine3dom: I don't think these are variables - they just annotate the format! macro. I personally hate the empty {} ones. Happy to remove if you want.

bovine3dom: https://doc.rust-lang.org/std/fmt/#named-parameters

src/project.rs
bovine3dom: I struggled with this a bit and got some horrible compiler errors. Could you point me in the right direction?

src/build/release.rs
synixebrett: I know they aren't variables outside the format, I just don't see a reason for their use in a simple string where each variable is in order and used once. Especially when they are using different names.

If rust ever adds string interpolation we would definitely use that, but for now, I think named parameters are not needed on simple strings like this.

src/project.rs
synixebrett: Unfortunately, serde default needs to be a function.

You can use something like [this](https://github.com/Dynulo/Gluon/blob/master/src/files/config.rs#L48), then we can add [this](https://github.com/Dynulo/Gluon/blob/master/src/files/config.rs#L54) to main.rs and use `crate::dft_false` as our default value. It isn't ideal, but it is better than an option.

src/build/release.rs
bovine3dom: It's because the format!/printf macros are alien to me. I find named parameters much more readable. I honestly found the line much easier to understand with it in; I only understood what it did after rewriting it.

If you feel as strongly but in the other direction I'll remove it.

synixebrett: #79 needs a little bit more testing, but I think that'll be a good compromise. There is no redeclaration, or like in your PR, declaring with different names. I am not a fan of changing the name for formatting, but I also think `name = name` looks ridiculous. This is the best we can do with hygienic macros unless rust one days adds string interpolation.

src/build/sign.rs
jonpas: Also not a fan.

bovine3dom:
Rebased, tidied up and all changes made. Tested and it still works (which almost never happens after I rebase).

src/build/release.rs
```diff
@@ -26,50 +29,83 @@ pub fn release(p: &crate::project::Project, version: &String) -> Result<(), Erro
         }
     }
 
-    // Generate key
     if !Path::new("releases/keys").exists() {
         fs::create_dir("releases/keys")?;
     }
+
+    let mut key;
```
synixebrett: I'm surprised the compiler allowed this, variables should always be given a value when they are defined. This should just be down on line 54.

bovine3dom: It's just a lazy way of allowing the two cases of `key` to store their result in a variable that the rest of the programme can read. If I put it down further, it'd go out of scope.

I guess the "right" way would be to do `let key = give_me_the_key(reuse_private_key: bool, ...)` and put the if below into it?

I don't really see the issue with how it is currently.

synixebrett: In rust `if` statements can return values. The if should return a value rather than using an empty value.

```rust
let x: int;
if something {
	case1();
	x = 1;
} else {
	case2();
	x = 2;
}
```
should be
```rust
let x = if something {
	case1();
	1
} else {
	case2();
	2
};
```

bovine3dom: Oh, that's magic. Thanks.

PR #52, opened 2019-03-08
Convert

synixebrett:
- Add `hemtt convert`

Requires #47

### DIFF
--- a/docs/utilities.md
+++ b/docs/utilities.md
@@ -23,3 +23,6 @@ Spanish          2015  75%
 Hungarian        1558  58%
 Turkish             3   0%
 ```
+
+# convert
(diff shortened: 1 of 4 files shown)


jonpas:
Convert Config might be a better name? Convert alone could mean a lot of things, I wouldn't think of converting configuration file as one of the first.

synixebrett:
Yeah, I original had `ConvertProjectFile`, but felt that was a bit long. `ConvertConfig` can work.

jonpas:
`ConvertConfig` could also mean Arma config, but that doesn't make that much sense anyways.

synixebrett:
I'll go with `ConvertProject` I think

PR #57, opened 2019-03-12
[WIP] Dependency graph

### LINKED ISSUE #54: Project dependency D(A?)G-maker
I'd like something that gives me the dependency graph of addons within a project. I think this probably is within the remit of HEMTT as a general ArmA-addon multi-tool.

It should be pretty straight-forward:

- check the requiredAddons[] of each config.cpp
- cross-reference with all the CfgPatches names (I guess we'd need to use armake for this?)
- work out what is external
- make the graph - probably using https://docs.rs/petgraph/0.4.13/petgraph/ and just exporting as a `dot` for now. I don't know how common my usage is so I'm not keen to build querying it into HEMTT.

My use-case is that I want to use the DAG to chop off bits of the graph that depend on external addons that I haven't white-listed and, on release, move these trees to separate mod-folder(s), merging on prefix.

Am I OK to go off and build this?

bovine3dom:
For some mad reason, I decided I wanted to do this properly. It therefore might take me a little while. I'm opening this PR mostly to let people keep an eye on it and tell me if I'm going very off-piste.

I've got up to the point where I have a big vector of all the processed `config.cpp`s.

I _could_ just do a text search for `CfgPatches` and `requiredAddons[]`, but since armake2 has to make a definition map anyway, it probably makes sense to try to get at that: https://github.com/KoffeinFlummi/armake2/blob/40fabd915514ffda372ec012b35ed4190d0e0515/src/preprocess.rs#L438

I'll probably do it the text search way first, as getting at that map without copy-pasting a large amount of code will probably require a PR to armake2.

Further down the to-do list is to ask armake2 to stop once it has got up to `CfgPatches.requiredAddons[]` in each file. (If someone defines `requiredAddons` more than once in a single config, they don't deserve my graphs anyway). I suspect that would speed stuff up a lot.

Closes #54 once finished.


notes.md
```diff
@@ -0,0 +1,17 @@
+# Making a dag out of deps
+
+## Runtime DAG
+
+Which mods can I split up into separate @ folders?
```
jonpas: Please don't write this, it's most annoying when you play on a joint op and the organizing community decided to rename or change folders around for no reason! 

Rather write it in the sense of checking which PBOs rely on which and what can be removed (especially in terms of modular mods).

bovine3dom: It's actually for our own modpack :) I want to make it more modular so I can debug stuff. It would be fun to do a `hemtt bisect` someday...

My current plan is just to output a DOT file. People can then do with that what they wish.

bovine3dom:
Mostly working now with the following caveats:

- Doesn't work with multiline requiredAddons
- need to trim the requiredAddons strings as apparently they include spaces sometimes
- need to send the addons to lowercase as they're apparently case insensitive

synixebrett:
You could try using [this](https://github.com/synixebrett/Config2JSON/blob/master/src/tojson.rs) to remove your regex, then use serde_json to turn it into a rust structure.

bovine3dom:
That would improve the current PR, but I'd like to eventually dig into `armake` to allow us to have early stopping, which would lead to invalid classes which I suspect your jsonifier wouldn't like. I suppose I should have a go at doing that first.

bovine3dom:
I mentioned the idea of `hemtt bisect` before, like `git bisect` but bisecting and descending trees until it found the folder that caused an error. We could similarly use it to create debug releases of a subset of addons and all of their dependencies. `hemtt build --recursive [addons]` seems like a sensible interface to me.

bovine3dom:
Added early stopping for a respectable speedup:

On a large project (~130 addons):

```
bash-5.0$ time ~/projects/HEMTT/target/debug/hemtt dependencies > test_master.dot

real    0m36.541s
user    1m40.267s
sys     0m0.397s
bash-5.0$ time ~/projects/HEMTT/target/debug/hemtt dependencies > test.dot

real    0m10.555s
user    0m35.824s
sys     0m0.143s
```

On a small project (with four addons):

```
bash-5.0$ time ~/projects/HEMTT/target/debug/hemtt dependencies > /dev/null

real    0m0.409s
user    0m0.996s
sys     0m0.007s
bash-5.0$ time ~/projects/HEMTT/target/debug/hemtt dependencies > /dev/null

real    0m0.487s
user    0m1.154s
sys     0m0.010s
```

Next stop - coaxing more information out of armake2's parser.

PR #60, opened 2019-03-18
Windows Installer

### LINKED ISSUE #26: Allow global installation and usage of HEMTT
Modern tooling are generally handed to developers as a user space or globally installed applications (eg. yarn, rustc, dotnet etc.) it would be a nice feature to consider to have HEMTT have this option too. The way I see it, there are two ways to approach this:

1. Create an installer that installs HEMTT binary and related dependencies if there ever will be some. The classic way.
2. Integrate an installer within HEMTT that will install the binary and related dependencies to the platform specific locations.

The implementation of both approaches have their own ups and downs.

### Separate installer
* Requires less custom implementation
* This would require a per platform installer or packager setup

### Integrated installer
* Needs to be implemented from scratch (Should be done and implemented as a separate package)
* Could provide a unified installation experience for all platforms: `./hemtt --install` & `hemtt --uninstall`
* Requires more knowledge to implement than separate installers
* Much more fun.

The reason why I'm not including scripts as an installer option is because they are basically the same as installer binaries / packages, but they're not very user friendly for the end user.

This, coupled with #25 would bring HEMTT up to par with most modern cli-based utilities.

synixebrett:
- Close #26 
- Add a windows installer to all tagged releases

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,7 +5,7 @@ authors = ["Brett <brett@bmandesigns.com>"]
 edition = "2018"
 
 [dependencies]
-armake2 = { git = "https://github.com/KoffeinFlummi/armake2", branch="master" }
+armake2 = { git = "https://github.com/KoffeinFlummi/armake2", branch = "master" }
 chrono = "0.4"
 colored = "1.7"
 docopt = "1"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -46,7 +46,7 @@ test_script:
       cargo build --target %TARGET% --release
     )
   # revert below before merge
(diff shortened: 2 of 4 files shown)


jonpas:
@ptMuta

jokoho48:
Appveyor should support Chocolatey so this should be able to install Inno Setup 5 with that
```
install:
  - choco install -y innosetup
  - set PATH="C:\Program Files (x86)\Inno Setup 5";%PATH%
```

synixebrett:
According to this page it should've been preinstalled https://www.appveyor.com/docs/windows-images-software/

PR #70, opened 2019-03-25
Fix sign pbo file prefix check

jonpas:
- Fix ACE3 `tracers` panic on release build

ACE3's `tracers` optional caused panic on release build, due to including the prefix `ace` (`trACErs`).

I am not sure this is the ideal solution, but I think it was intended to check only start anyways.


bovine3dom:
I have put this in #51 manually because I didn't want to forget about it - I'll move everything around so much that it would probably become a pain to merge.

jonpas:
No worries @bovine3dom just close this when it's merged then! :)

jonpas:
Actually, can we get this merged, so I can work on other ACE3 stuff before #51 gets merged? Looks like there is still quite a discussion there.

PR #72, opened 2019-03-28
Document convertproject on help

bovine3dom:
`convertproject` wasn't listed under `--help`


src/main.rs
```diff
@@ -65,6 +65,7 @@ Commands:
     build               Build the project
     clean               Clean build files
     update              Update HEMTT
+    convertproject      Convert hemtt.json to hemtt.toml
```
jonpas: Convert between the two.

PR #74, opened 2019-03-28
Add Git Template Helper

### LINKED ISSUE #32: Template Helper - Git
Utility that retrieves current git commit hash:
- from `git` binary if available
- from `.git` folder if available

Ref.: [make.py](https://github.com/acemod/ACE3/blob/191c4fb694b80337c35a36be369b62c634f0e079/tools/make.py#L693)

Should support retrieving it as truncated (to N characters) for use in configurable options.

jonpas:
- Add Git Template Helper - close #32
  - Use [`git2`](https://docs.rs/git2/0.8.0/git2/) (includes `libgit2-sys` so no extra git installation is required)
- Add `id <number>` parameter to get SHA-1 of last commit and truncate to given number of characters
- Add documentation for Git Template Helper
- Fix `modname` ignoring templating

Could use some idea for formatting parameters, maybe to make them universal for other possible template helpers as well. Otherwise I am satisfied with this for the current limited use.


docs/templating.md
```diff
@@ -26,3 +26,13 @@ Date can be used to get information about the current date and time. HEMTT uses
 ```
 {{date \"%y%m%d\"}}
 ```
+
+## git
+Git helper can be used to get information about the git repository.
+
+Parameters:
+- `id <number>`: id (SHA-1) of HEAD revision _(number from 1 to 40, default: 8)_
```
bovine3dom: It could just be that I'm an idiot, but this to me reads as if you're going to give me the SHA of `HEAD~8`. I'm guessing it's actually truncated, so I'd just say `... HEAD truncated to _..._ characters`.

jonpas:
```suggestion
- `id <number>`: id (SHA-1) of HEAD revision truncated to `<number>` characters _(from 1 to 40, default: 8)_
```

synixebrett:
What scenarios has this been tested under? Window / Linux? Will it work without git installed?

jonpas:
Linux with HEMTT-Example. I didn't get around to testing it on Windows and without git yet, it should work as long as there is a .git folder

PR #75, opened 2019-03-28
Clear keys when clearing releases

jonpas:
- Title


src/files.rs
```diff
@@ -42,6 +42,10 @@ pub fn clear_release(version: &String) -> Result<(), Error> {
         println!("  {} old release v{}", "Cleaning".yellow().bold(), version);
         fs::remove_dir_all(format!("releases/{}", version))?;
     }
+    if Path::new("releases/keys").exists() {
+        println!("  {} old keys", "Cleaning".yellow().bold());
+        fs::remove_dir_all("releases/keys")?;
```
bovine3dom: This would delete all of the keys ever - the private key if it exists, and all keys for versions other than the one being cleared. That would be surprising behaviour to me. I'd expect it only to remove keys associated with the release I'm trying to release.

jonpas: I tend to agree, but I was lazy, need to find the correctly named one here too. ^^

jonpas:
```rust
Path::new(&format!("releases/keys/{}", p.get_keyname()))?;
```
I guess.

bovine3dom: But won't it just be overwritten anyway?

jonpas: Apparently not.

synixebrett:
This should wait for #51, then have those TODOs completed.

jonpas:
@bovine3dom does this work for you?

PR #79, opened 2019-03-30
Macros

synixebrett:
Create macros allowing for faux string interpolation without redundant declaration like named parameters use.

```rust
let name = "Brett";
iprintln!("Hello {name}", name);
```
compared to the rust
```rust
let name = "Brett";
println!("Hello {name}", name = name);
```


synixebrett:
Added a few more macros.

**Colors**
```rust
yellow!("Testing", "This is a test");
blue!("test", "This is yet another test");
```


**Progress Bar**
```rust
pbm.lock().unwrap().pb().finish_print(&format!("  {} {}{}", "Executed".green().bold(), built.len(), crate::repeat!(" ", 60)));
println!()
```
is now
```rust
crate::finishpb!(pbm.lock().unwrap().pb(), green, "Executed", built.len());
```

PR #94, opened 2019-04-04
Try to improve errors on dodgy JSON

bovine3dom:
If there's an error in hemtt.json (or toml, presumably?), you get a scary error like:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Custom { kind: InvalidData, error: Error("expected `:`", line: 4, column: 11) }', src/libcore/result.rs:1009:5
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

With the first commit in this PR, you get the slightly friendlier

```
error:  expected `:` at line 4 column 11
```

and with the second:

```
error: Could not parse project configuration: expected `:` at line 4 column 11
```

It still isn't ideal - it should really include the file name - but I think this is a worthwhile step in the right direction.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -5,6 +5,7 @@ use std::fmt::{Debug, Display};
 pub trait ErrorExt<T, E> {
     fn print(self) -> Option<T>;
     fn unwrap_or_print(self) -> T;
+    fn unwrap_or_print_with_context(self, context: &str) -> T;
 }
 impl<T, E: Debug + Display> ErrorExt<T, E> for Result<T, E> {
     fn print(self) -> Option<T> {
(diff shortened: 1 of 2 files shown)


src/error.rs
```diff
@@ -14,11 +15,14 @@ impl<T, E: Debug + Display> ErrorExt<T, E> for Result<T, E> {
         }
         Some(self.unwrap())
     }
-    fn unwrap_or_print(self) -> T {
+    fn unwrap_or_print_with_context(self, context: &str) -> T {
```
bovine3dom: If anyone knows a better way of having optional arguments, I'm all ears :)

src/error.rs
```diff
@@ -14,11 +15,14 @@ impl<T, E: Debug + Display> ErrorExt<T, E> for Result<T, E> {
...
         if let Err(error) = &self {
-            eprintln!("{}: {}", "error".red().bold(), error);
+            eprintln!("{}: {}{}", "error".red().bold(), context, error);
```
bovine3dom: It's probably worth doing

```
if context != "":
    context += " "
```

but I ran out of time for this PR and I know that Rust wouldn't make it easy.

synixebrett:
I agree that this needs to be improved, but I wasn't happy with how this PR was going to solve it. I have been wanting to find a better solution. I still think there should be a better way of handling errors, so for now I won't merge this PR.

synixebrett:
This is improved in #140

PR #108, opened 2019-05-01
default to v2 signatures

synixebrett:
- Use V2 Signatures by default
- Add project level setting `sigversion`


bovine3dom:
I think it'd be sensible to document why we're using V2 and not V3.

jonpas:
It's only temporary.

bovine3dom:
There's nothing more permanent than the temporary :)

PR #132, opened 2019-06-14
Support Launchable / Folder Optionals

jonpas:
- Build (by default) optionals into their own mod folders, allowing users to directly launch them.
- Cleanup `fn release` a bit
  - Base release folder variable
  - Filtering for files only
  - Pass path directly to `copy_sign`

Used in ACE3 and other projects - easier to do in HEMTT than custom build scripts, also makes more sense. Enabled by default, possible to disable.

TODO:
- [x] Documentation
- [x] Windows compatibility test

### DIFF
--- a/docs/project.md
+++ b/docs/project.md
@@ -132,6 +132,15 @@ HEMTT will build the specified addons from the `./optionals` folder.
 ]
 ```
 
+## folder_optionals
+**Type**: Bool
+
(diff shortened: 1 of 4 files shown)


jonpas:
Am open to different name if anyone has better suggestions.

jonpas:
Confirmed working on Windows, documentation added.

PR #133, opened 2019-06-14
Optimize zip utility (2x speed increase)

jonpas:
- Optimize zip utility for 2x speed increase using:
  - `BufReader` and `BufWriter`
  - Direct copying without buffer

Test results on 2 of our random mods with `debug` build:

Old   | New  | Zip Size
----- | ---- | --------
28 s  | 14 s | 11.2 MB
122 s | 62 s | 50.6 MB

Deflate implementation in zip-rs is still terribly slow. Bzip2 is very fast, but does not natively work on Windows.

### DIFF
--- a/src/utilities/zip.rs
+++ b/src/utilities/zip.rs
@@ -6,7 +6,7 @@ use walkdir;
 use zip;
 
 use std::fs::File;
-use std::io::{Read, Write, Error};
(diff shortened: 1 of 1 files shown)


jonpas:
I just did something way better, now getting 7z-like speeds. Will be in another PR. :)

jonpas:
Replaced by #134.

Issue #66, opened 2019-03-23
Addon function tools


jonpas:
Create, delete, rename and move a function inside a specific addon.
Should support templating as well.

Done in Python in https://github.com/acemod/ACE3/pull/4631, but HEMTT would be ideal for this, building on top of `hemtt addon`.

PR #140, opened 2019-06-19
Restructure

synixebrett:
I have been unhappy with the structure of HEMTT for a while now since it started as a weekend project on holiday and more features just got hacked into it as they were needed. Now that I know what I actually want HEMTT to accomplish I'm restructuring the core of the program to avoid more hacky methods.

- [ ] Template Management
  - [x] Addons
  - [ ] Functions
- [ ] Improved Error Handling
  - [x] More readable output
  - [x] Preprocessor
  ~- [ ] Linting~ Moved to #170 
  - [ ] Non-Standard Practice
    - [x] addons with prefix
    - [x] addons with non A-z names


jonpas:
Can you describe what this changes for the use of the program? What is LUA for?

synixebrett:
It will be used for managing templates.
Adding a feature like #66 is complicated since there are many different ways functions could be organized in the project. Those features will be implemented via Lua like the WIP [cba template](https://github.com/hemtt/cba/tree/master/scripts)

jonpas:
Sounds good! Not really a fan of LUA, but it's probably simplest to use here.

Let me know when this is ready for testing.

docs/project.md
```diff
@@ -173,8 +173,8 @@ The default is set according to the following table:
 
 | `reuse_private_key` value | Default `keyname`         |
 | ------------------------- | ------------------------- |
-| `false`                   | `{{prefix}}_{{version}}`  |
-| `true`                    | `{{prefix}}`              |
+| `false`                   | `{{prefix}}}_{{version}}`  |
```
jonpas: Triple `}}}`? :thinking:

synixebrett: Find all and replace mistake, fixed in next commit

src/startup.rs
```diff
@@ -0,0 +1,47 @@
...
+    exec!(deprecated);
+}
+
+/// Checks for the recommended items in a .gitignore
+/// Display a warning if they are not found
+fn check_git_ignore() -> Result<(), HEMTTError> {
+    if Path::new(".gitignore").exists() {
+        let mut data = String::new();
+        open_file!(".gitignore")?.read_to_string(&mut data)?;
+        let mut ignore = vec!("releases/*", "*.biprivatekey");
```
jonpas: Private key is in releases anyways.

.editorconfig
```diff
@@ -1,7 +1,7 @@
 root = true
 
 [*]
-end_of_line = lf
+end_of_line = crlf
```
jonpas: What? Why?

synixebrett: It's just temporary since something was complaining on my machine and I'm too lazy to fix right now. `lf` is still going to be used

src/startup.rs
synixebrett: https://github.com/synixebrett/HEMTT/blob/50eb56845eb3914b9659355a2d8aa6a94e7c497a/src/commands/build/postbuild/sign.rs#L25

The keys should be backed up somewhere anyway, but this will try to prevent people from losing their private key. I was going to inquire about adding `keys/` too

synixebrett: Since the only time the private key would be written to disk, is if they want to reuse it.

jonpas: Yeah, makes sense.

.editorconfig
jonpas: Alright. Yeah, entire repo is forced to LF, so when you push it changes anyways. But you should fix your editor or local git. :D

PR #165, opened 2019-08-02
Enable macOS builds on Travis

Dahlgren:
- Enable macOS builds on Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@
 
 dist: trusty
 language: rust
-services: docker
 sudo: required
 
 env:
@@ -49,8 +48,8 @@ matrix:
     # - env: TARGET=s390x-unknown-linux-gnu DISABLE_TESTS=1
 
     # OSX
-    # - env: TARGET=x86_64-apple-darwin
-    #   os: osx
+    - env: TARGET=x86_64-apple-darwin
+      os: osx
     # - env: TARGET=i686-apple-darwin
     #   os: osx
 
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -3,8 +3,6 @@
 
 dist: trusty
 language: rust
-services: docker
-sudo: required
```
Dahlgren: it's not really needed right?

.travis.yml
```diff
@@ -3,8 +3,6 @@
 
 dist: trusty
 language: rust
-services: docker
```
Dahlgren: docker is not supported on macOS and not used on Linux as far as I can tell

ci/version.sh
```diff
@@ -1,5 +1,5 @@
 if [ -z "$TRAVIS_TAG" ]
 then
     HASH=$(git log --pretty=format:'%h' -n 1)
-    sed -i -E "s/version = \"([^\"]+?)\"$/version = \"\1-$HASH\"/" Cargo.toml
+    sed -E "s/version = \"(.+)\"$/version = \"\1-$HASH\"/" Cargo.toml > Cargo.toml.versioned && mv Cargo.toml.versioned Cargo.toml
```
Dahlgren: GNU sed and BSD sed works differently, especially the `-i` argument. AFAIK this is the simplest way to replace content inside the file being read, writing stdout directly to it sadly results in an empty file

PR #199, opened 2019-09-27
Utility - File Patching

### LINKED ISSUE #14: Symlink project to Arma for filepatching
To facilitate immediate and easy file patching setup.

Could be ran manually with command or automatically setup on each build if it doesn't exist yet, not sure which approach is better, automatic seems a bit intrusive.

synixebrett:
- Close #14


synixebrett:
Originally `remove_file!` was part of this PR, but then it's use got removed. I know it's now unrelated to this PR but I'm OK with that given how minor this PR is.

Issue #253, opened 2020-01-23
Single PBO Mods without Prefix File-Patching Support


jokoho48:
**HEMTT Version:** `0.7.3` (stable or dev + commit hash)
**Project:** https://github.com/TaktiCool/CLib/commit/fa3cc40543401d2aee06f7865532de9803e2ab3c

**Description:**

Add the Possibility to use FilePatching Command without having a Prefix Defined, Some mod Structures may don't need Prefixes or multiple PBOs or use then.
and when I would define a Prefix my mod would build to Prefix_Prefix what would not be the intended filename

My idea would be if the Prefix is not defined or an Empty string to simply fall back to the current Folder Name

**Steps to reproduce:**

without Prefix
- Try to create file-patching version of linked commit of the CLib Project
- Link created at ARMAFOLDER\\MAINPREFIX\\

- Expected is a Link Created at ARMAFOLDER\\MAINPREFIX\\FOLDERNAME

with Prefix
- On Build the pbo is Builded with the name CLib_CLib and the message 
`warning: Redundant prefix in addon name "CLib"
    use "", pbos are prefixed automatically`

**HEMTT Output:**
```
$ .\hemtt.exe file-patching
HEMTT 0.7.3
Environment: dev

Linked at "E:\\\\Programme\\\\SteamLibrary\\steamapps\\common\\Arma 3\\tc\\"
You can now use `-mod="\tc\" -filePatching`
```

PR #258, opened 2020-01-25
add support for Prefixed named PBOs

jokoho48:
- possible fix for #253

### DIFF
--- a/src/commands/build/addon.rs
+++ b/src/commands/build/addon.rs
@@ -46,7 +46,7 @@ impl Addon {
 
     pub fn target(&self, p: &Project) -> PathBuf {
         let mut target = PathBuf::from(self.location.to_string());
-        if !p.prefix.is_empty() {
+        if !p.prefix.is_empty() && self.name != p.prefix {
             target.push(&format!("{}_{}.pbo", p.prefix, self.name));
         } else {
             target.push(&format!("{}.pbo", self.name));
@@ -72,7 +72,6 @@ impl Addon {
             r.push(&format!("@{}_{}", p.modname().unwrap(), self.name));
             r.push("addons");
         }
-
         r
     }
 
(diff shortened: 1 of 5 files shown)


synixebrett:
Ah ok I see how you did this, I'd personally be more in favor of supporting no prefix, rather than making the prefix the same as the PBO name. Then it would also support two PBOs being present without a prefix.

jokoho48:
I think both should be possible. but this was for me now the easier one to find and fix

jokoho48:
I think I found a fix for the Filepatching after some digging around

src/commands/build/addon.rs
```diff
@@ -46,7 +46,7 @@ impl Addon {
 
     pub fn target(&self, p: &Project) -> PathBuf {
         let mut target = PathBuf::from(self.location.to_string());
-        if !p.prefix.is_empty() {
+        if !p.prefix.is_empty() && self.name != p.prefix {
```
synixebrett: Not a fan of these checks, the prefix should either be defined or not set

src/project/signing.rs
```diff
@@ -25,7 +25,11 @@ impl Project {
 
     /// Get the name for .bisign files
     pub fn get_sig_name(&self, pbo: &str) -> Result<String, HEMTTError> {
```
synixebrett: This should have a check for no prefix instead of checking if the prefix is the pbo name

src/utilities/filepatching.rs
```diff
@@ -13,21 +13,22 @@ impl FilePatching {
         let mut dir = dir;
         dir.push(&p.mainprefix);
         let mut target = dir.clone();
-        target.push(&p.modname()?);
+        let modname = &p.modname()?;
```
synixebrett: I don't see what any of the changes to this file accomplish logic wise

src/commands/build/addon.rs
jokoho48: should we maybe then add a PathPrefix and a Prefix? because we in some cases need a prefix for the file-patching command

src/utilities/filepatching.rs
jokoho48: Somehow when I don't store it as a variable it changes midway and i could not find out why. this was the only was I could make it work

src/commands/build/addon.rs
synixebrett: File patching should not require a prefix, it only needs the `mainprefix` and the `modname`

src/project/signing.rs
jokoho48: good point will change that

src/commands/build/addon.rs
jokoho48: file patching was current build around the prefix that was the issue I have run into.

PR #273, opened 2020-02-03
Fix too broad configuration file search

### LINKED ISSUE #267: HEMTT binary used as configuration file if same name
**HEMTT Version:** `0.7.4` (stable)
**Project:** Any

**Description:**

Having a Linux HEMTT binary `hemtt` in project root next to configuration file `hemtt.toml` causes HEMTT to find itself and fail instead of finding the proper file with the extension.

**Steps to reproduce:**

- Name HEMTT binary `hemtt` and place it next to `hemtt.toml` configuration file.
- Try building the mod.

**Additional information:**

N/A

**HEMTT Output:**

```
$ ls
addons hemtt* hemtt.toml ...
$ hemtt build
configuration file "/home/jonpas/Data/Work/Arma 3/Mods/TheseusServices/hemtt" is not of a registered file format
```

jonpas:
- Fix #267 

Not pretty, it works. Could setup a nicer data store to check both, but there is no plan to support anything else.

### DIFF
--- a/src/project/mod.rs
+++ b/src/project/mod.rs
@@ -149,9 +149,6 @@ impl Project {
         if Path::new("hemtt.toml").exists() {
             // Single file (toml)
             p.merge(File::with_name("hemtt.toml").required(true))?;
-        } else if Path::new("hemtt.json").exists() {
-            // Single file (json)
-            p.merge(File::with_name("hemtt.json").required(true))?;
         } else {
             // Project folder
             if !Path::new(".hemtt/").exists() {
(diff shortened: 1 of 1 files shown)


src/project/mod.rs
```diff
@@ -146,9 +146,12 @@ impl Project {
         debug!("Root Directory: {:?}", root);
         std::env::set_current_dir(root)?;
 
-        if Path::new("hemtt.toml").exists() || Path::new("hemtt.json").exists() {
-            // Single file
-            p.merge(File::with_name("hemtt").required(true))?;
+        if Path::new("hemtt.toml").exists() {
+            // Single file (toml)
+            p.merge(File::with_name("hemtt.toml").required(true))?;
+        } else if Path::new("hemtt.json").exists() {
```
synixebrett: Since JSON is deprecated I don't think this needs to be here

PR #295, opened 2020-03-14
Support directories in files configuration

jonpas:
**When merged this pull request will:**
- Finish all set goals of #88
- Support copying directories specified in `files` in 2 ways:
  - Specified folder into root of release folder
    - Example: `"extras/plugin` -> `releases/<version>/@<mod>/plugin`
  - Specified folder with mirrored structure (`/` at the end)
    - Example: `"extras/plugin/` -> `releases/<version>/@<mod>/extras/plugin`

PR #303, opened 2020-03-17
Support file directory mirror in files configuration

jonpas:
- Expand #295 for files (not just directories)
- Allow specifying file configuration as such:
  - Specified file into root of release folder
    - Example: `"extras/test.txt` -> `releases/<version>/@<mod>/test.txt`
  - Specified file with mirrored directory structure (`/` at the end)
    - Example: `"extras/test.txt/` -> `releases/<version>/@<mod>/extras/test.txt`

Required for ACRE2.

### DIFF
--- a/src/commands/build/postbuild/release.rs
+++ b/src/commands/build/postbuild/release.rs
@@ -51,17 +51,22 @@ impl Task for Release {
             })?;
         }
 
-        for file in &p.files {
-            for entry in glob(file)? {
+        for mut file in p.files.to_owned() {
+            let mut mirror_structure = false;
+            if file.ends_with('/') {
+                file.pop();
+                mirror_structure = true; // Mirror directory structure if path ends in slash
+            }
+
+            for entry in glob(&file)? {
                 if let Ok(path) = entry {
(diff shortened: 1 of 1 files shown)


jonpas:
Give me another way then, that won't interfere with path. I already documented this way in the previous PR.

jonpas:
For reference, if I do find out I need from -> to hash map for files, I'll add it. And it can easily be side-by-side.

src/commands/build/postbuild/release.rs
```diff
@@ -51,17 +51,24 @@ impl Task for Release {
...
-        for file in &p.files {
-            for entry in glob(file)? {
+        for mut file in p.files.to_owned() {
+            // Mirror directory structure if path ends in slash
+            let mirror_structure = if file.ends_with('/') {
+                file.pop();
+                true
+            } else {
+                false
+            };
```
jokoho48: why not like this?
```suggestion
            let mirror_structure = file.ends_with('/');
            if mirror_structure {
                file.pop();
            };
```

PR #400, opened 2021-01-04
Trim leading and trailing slashes from pbo prefix header

Dahlgren:
- Behave like Mikero Tools
- Save me from wasting time on troubleshooting why CfgFunctions won't load properly


Dahlgren:
The headers logic should probably be refactored a bit and be given proper tests

BrettMayson:
Included in experiment branch
https://github.com/BrettMayson/HEMTT/blob/d1082b3d42ec0e763760bcbea99741f47134868b/hemtt-pbo/src/pbo/writing.rs#L87

PR #466, opened 2023-02-20
create `.hemtt` before creating `.hemtt/project.toml`

### LINKED ISSUE #465: `hemtt new` results in path not found error
Running `.\hemtt.exe new somemod`. Entering all the data. After selecting a license and pressing enter, I get an error: `error: IO Error: The system cannot find the specified path. (os error 3)`.

I'm on Windows 10, running the latest alpha build (a17).

I've not used rust, but presumably
https://github.com/BrettMayson/HEMTT/blob/8545e628401a1f143595c213a23d384c30892a89/bin/app/src/commands/new.rs#L52 attempts to create `.hemtt/project.toml` without creating the `.hemtt` subfolder first.

https://doc.rust-lang.org/std/fs/struct.File.html#method.create says "Depending on the platform, this function may fail if the full directory path does not exist.". I guess Windows is one of those.

Chronophylos:
fixes #465

### DIFF
--- a/bin/app/src/commands/new.rs
+++ b/bin/app/src/commands/new.rs
@@ -49,7 +49,9 @@ pub fn execute(matches: &ArgMatches) -> Result<(), Error> {
     git2::Repository::init(path)?;
 
     // Create .hemtt/project.toml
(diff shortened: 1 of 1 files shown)


BrettMayson:
Thanks for the contribution!

Issue #338, opened 2020-05-18
Preprocessing on UNC paths very slow


jonpas:
**HEMTT Version:** `0.7.6` (stable
**Project:** Any

**Description:**

Building mods on UNC drives is considerably slower, simple observation shows `Preprocess` steps is the culprit, everything else is about the same.

**Steps to reproduce:**

- Have mod mounted on UNC drive
- Build the mod

**Additional information:**

I will try to do some profiling.

**HEMTT Output:**

/

BrettMayson:
Can this be closed? Likely not anything HEMTT could do anyway, maybe buffered vs unbuffered writes somewhere but I think this is pretty niche.

jonpas:
Am yet to test this on HEMTT v1.

jonpas:
Can't actually test this because it's failing to make links between temporary directory and my UNC path (network drive).

```sh
TRACE link "C:\Users\Jonpas\AppData\Local\Temp\hemtt\Arma 3_Mods_TAC_Mods\source\x\tac\addons\apollo" => "E:\Arma 3\Mods\TAC_Mods\addons\apollo"
error: Unable to create link: Local volumes are required to complete the operation.
```

This happens because `mklink /J` (Junction) is used. `mklink /D` (Directory symbolic link) would work correctly, maybe fall-back to that could be added?

jonpas:
With #491 I could test this and I don't see any slow-down like I did with the old HEMTT.

1min 31s for full ACE3 build with binarization on 8c/16t inside a Windows VM.

PR #491, opened 2023-04-17
Fall-back link creation to directory symbolic link

jonpas:
Fix compatibility with network drives (UNC paths). Related to #338.

Maybe can just do directory symbolic link at all times? They mostly behave the same.


jonpas:
Confirmed working on my machine/setup.

PR #523, opened 2023-06-15
Support --no-archive in rename zip example hook

jonpas:
There might be a better way to do this? Maybe we could get `is_archive()` or something, that way we could skip a hook if the flag was actually given, and still get a proper error if the flag was not given and zip also didn't exist.


jonpas:
`src.is_file()` where `src` is some sub-path of `HEMTT_RFS` actually doesn't seem to function either, I get a crash:
```
ERROR panicked at 'called `Result::unwrap()` on an `Err` value: ErrorFunctionNotFound("is_file (std::path::PathBuf)", 6:8)', bin/src/modules/hook/mod.rs:150:51
```

BrettMayson (approved):

jonpas:
Don't merge it, it doesn't actually work (see comment before this one).

BrettMayson:
Ye, I was just going to merge this after I get a change to investigate that, it'd be another PR fixing it

BrettMayson:
Root issue fixed in f05f24e99b71e0200fb4d6e6294a0ff3ae7cdda0

PR #549, opened 2023-08-20
Preprocssor - Warn if `#if`s not closed

### LINKED ISSUE #548: Should error if macro not closed
```
#ifdef A
    #ifdef B
    x = 1;
    #else
    x = 2;
#else
    #ifdef B
    x = 3;
    #else
    x = 4;
#endif
```
subtle error in this code, but hemtt will build with no `x=` present, 
should probably error if `#if` macros aren't closed

PabstMirror:
Close #548

### DIFF
--- a/libs/preprocessor/src/ifstate.rs
+++ b/libs/preprocessor/src/ifstate.rs
@@ -23,6 +23,10 @@ impl IfStates {
         Self(Vec::new())
     }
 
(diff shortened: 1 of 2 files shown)


BrettMayson:
Going to implement this in #546 as a proper error rather than just a warning printed to stderr

Issue #561, opened 2023-09-21
[Feature Request] hemtt.launch - start into editor mission-parameter


PulsarNeutronStar:
Based on https://community.bistudio.com/wiki/Arma_3:_Startup_Parameters

>`"<path>\mission.sqm"` | Load a mission directly in the editor. Example: "c:\arma3\users\myUser\missions\myMission.intro\mission.sqm"

a parameter like 

```
[hemmt.launch.test]
parameters = [
    "C:\Users\myUser\Documents\Arma 3 - Other Profiles\ProfileName\missions\test.Stratis\mission.sqm",
    "-skipIntro",           # These parameters are passed to the Arma 3 executable
    "-noSplash",            # They do not need to be added to your list
    "-showScriptErrors",    # You can add additional parameters here
    "-debug",
    "-filePatching",
    "-window",
]
```

should work, but instead, when launching, hemtt throws the following issue

```
ERROR Toml Error: TOML parse error at line 60, column 10
   |
60 |     "C:\Users\myUser\Documents\Arma 3 - Other Profiles\ProfileName\missions\test.Stratis\mission.sqm",
   |          ^
invalid unicode 8-digit hex code
```

BrettMayson:
Need to use double `\\`

PulsarNeutronStar:
for each? - might be worth putting in the hemtt book as an example then :)

PulsarNeutronStar:
Can confirm, `\\` does the trick!

PR #563, opened 2023-09-23
[Hemtt Book] added Example for "Launch Editor Mission" Parameter

PulsarNeutronStar:
See #561


BrettMayson:
I want to discourage anything that would be specific to someone's machine, updated the example

PR #574, opened 2023-10-14
Config - Warn on CfgMagazineWells not found in CfgMagazines

BrettMayson:


PabstMirror:
cba_jam defines some magwells with vanilla mags
https://github.com/CBATeam/CBA_A3/blob/master/addons/jam/magwells_556x45.hpp#L21-L22

BrettMayson:
It only checks magazines that start with the project's prefix

PR #591, opened 2023-10-30
Binarize: fix copy files from addons folder path 

jonpas:
Fix path from #590. Tested only on CUP, ~~it might go a folder too high in other mods? That would still work though~~.

### DIFF
--- a/bin/src/executor.rs
+++ b/bin/src/executor.rs
@@ -144,7 +144,7 @@ fn setup_tmp(ctx: &Context) -> Result<(), Error> {
         }
         let tmp_file = tmp
             //.join(ctx.addons().first().unwrap().prefix().as_pathbuf())
-            .parent()
+            //.parent()
             .unwrap()
             .join(file.file_name().unwrap());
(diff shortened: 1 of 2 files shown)


BrettMayson:
Using the prefix and then just parenting it if it, is it not better to just remove the prefix and patent?

jonpas:
> Using the prefix and then just parenting out of it, is it not better to just remove the prefix and parent?

That results in `%tmp%/hemtt/source/<file>` instead of `%tmp%/hemtt/source/<mod>/<file>`.

BrettMayson:
That'd still work though, at least for model.cfg

I just want to make sure we avoid another issue of the prefixes being weird and it ends up somewhere inaccessible

jonpas:
> That'd still work though, at least for model.cfg
> 
> I just want to make sure we avoid another issue of the prefixes being weird and it ends up somewhere inaccessible

Correct, it did work. I will adjust then.

BrettMayson (approved):

PR #597, opened 2023-11-02
Config: __eval support

BrettMayson:


jonpas:
Ref. https://github.com/KoffeinFlummi/armake2/issues/23#issuecomment-504057928 or it will get lost again. :grimacing:

PR #649, opened 2024-02-07
Improve support for direct mission launching

### LINKED ISSUE #621: Add option to use realtive path for mission file with launch
Add an option to use a relative path mission with the launch option.
This will allow to have a test mission in the git repo, allowing for easier testing between multiple developers.

Not sure if even possible from Arma side, if it doesn't require all missions to be in the documents/profile folders.

mharis001:
Improve support for directly launching into 3DEN with an existing mission by allowing missions contained in the project to be used.

Adds a `mission` configuration parameter that specifies whether a launch configuration loads directly into a mission. Can be specified as either:
  - a folder name in `.hemtt/missions/` (e.g., `mission = "test.VR"` would launch `.hemtt/missions/test.VR/mission.sqm`).
  - a relative path to a `mission.sqm` file (or folder containing it).

Tested all of the various cases and everything looks good. I am not too familiar with Rust and this is just the result of my messing around. So, the code could probably be improved.

Close #621

### DIFF
--- a/bin/src/commands/launch/mod.rs
+++ b/bin/src/commands/launch/mod.rs
@@ -283,24 +283,26 @@ pub fn execute(matches: &ArgMatches) -> Result<Report, Error> {
 
         if !path.is_absolute() {
             path = std::env::current_dir()?.join(mission);
-        }
 
-        if !path.ends_with("mission.sqm") {
-            path.push("mission.sqm");
-        }
+            if !path.ends_with("mission.sqm") {
+                path.push("mission.sqm");
+            }
 
-        if !path.is_file() {
-            path = std::env::current_dir()?
-                .join(".hemtt")
-                .join("missions")
-                .join(mission)
-                .join("mission.sqm");
-        }
+            if !path.is_file() {
+                path = std::env::current_dir()?
+                    .join(".hemtt")
+                    .join("missions")
+                    .join(mission)
+                    .join("mission.sqm");
+            }
 
-        if path.is_file() {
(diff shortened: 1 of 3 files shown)


bin/src/commands/launch/mod.rs
```diff
@@ -278,6 +278,32 @@ pub fn execute(matches: &ArgMatches) -> Result<Report, Error> {
...
+        if !path.is_file() {
+            path = std::env::current_dir()?
+                .join(".hemtt")
+                .join("missions")
+                .join(mission)
+                .join("mission.sqm");
+        }
+
+        if path.is_file() {
+            args.push(format!("\"{}\"", path.display().to_string()));
```
BrettMayson, line 301:
```suggestion
            args.push(format!("\"{}\"", path.display()));
```

bin/src/commands/launch/mod.rs
```diff
@@ -278,6 +278,32 @@ pub fn execute(matches: &ArgMatches) -> Result<Report, Error> {
             .join(" "),
     );
 
+    if let Some(mission) = launch.mission() {
+        let mut path = PathBuf::from(mission);
+
+        if !path.is_absolute() {
+            path = std::env::current_dir()?.join(mission);
+        }
```
BrettMayson, line 286: I would probably rather intentionally not allow absolute paths, and only missions in the .hemtt/missions folder or elsewhere in the project

BrettMayson (changes requested):

BrettMayson (approved):

Issue #596, opened 2023-11-02
Textures from Arma 3 not used during binarization


jonpas:
Models that reference textures from Arma 3's content `\a3\...` currently do not get binarized correctly, as HEMTT does not pass those textures to binarize. A good example is any transparency on models, such as glass on optics.

HEMTT should allow using referenced textures and potentially warn about missing ones.

---

This is also a deeper problem as we do not want to copy textures to `include` folder due to licensing/copyright restrictions and the potential size.

BrettMayson:
I'm sure we'll be able to get the includes going, but yeah, the copyright will kinda make the lack of a P drive a problem here.

If you have a P drive setup, it works with HEMTT as-is, right?

jonpas:
> If you have a P drive setup, it works with HEMTT as-is, right?

It does not.

jonpas:
> [02:19]jonpas: Ok I actually have some useful information. It was in fact just a matter of copying some rvmats and textures used in the scope models, to `include/a3/...`
[02:20]jonpas: So the real fix to avoid copying a ton of files for content mods, is allow them to still use P-drive. pboProject copies all textures in use to a `P:\temp` folder. I wonder if we could just symlink a P-drive to HEMTT temp if enabled/required in project config. 
[02:21]Brett: I wonder how much CUP actually needs from the P drive
[02:21]jonpas: probably not that much tbf
[02:22]Brett: It'd be neat to check the files with HEMTT, see what it needs, and grab those from the regular Arma install
[02:22]Brett: no need for a massive unpack of everything
[02:22]jonpas: But there is no way to check what is required without going through each item and attachment in-game, and even then it might not be obvious :/
[02:22]Brett: ?
[02:22]jonpas: binarize doesn't complain if textures are missing
[02:22]Brett: wouldn't all the files needed be found in the .p3ds?
[02:23]jonpas: p3ds or rvmats in use by the p3d, or rvmats in use by the rvmats in use by the p3d... etc 😛 
[02:23]jonpas: But yeah, if we can log "this file is missing for proper binarization", it would basically solve the problem
[02:24]jonpas: pboProject does that and that's the only thing stopping binarization; binarizer itself does no file checking at all

---

PboProject also binarizes whole component, no each file separately, see the binarize calls:
```sh
HEMTT Call
["<Steam>\\Arma 3 Tools\\Binarize\\binarize_x64.exe",
"-norecurse",
"-always",
"-silent",
"-maxProcesses=0",
[...]

PR #667, opened 2024-03-14
p3d, missing textures & materials

BrettMayson:
```toml
[hemtt.build]
pdrive = "require"
```
1. Will use a mounted P:\ drive first if one exists
2. Will look for a setup but unmounted P drive from the Arma 3 Tools settings
3. Will use an Arma 3 steam installation and extract needed files to %temp%\hemtt


BrettMayson:
@mike-mf for testing too please

BrettMayson:
ok I should be done pushing this now :| fixed the linux issue very efficiently in 3 commits

jonpas:
Mike has I think, I'll try to do it today.

BrettMayson:
Going to merge as I want to make some internal changes on main and don't feel like dealing with a merge later, I think this should be good to go. We can use the issue #596 to finish confirming this

PR #712, opened 2024-06-19
config: Check if any classes are duplicated

PabstMirror:
Catches any duplicate class
e.g.
```
class a;
class a {};

class b {};
class b {};
```


BrettMayson:
```
class a;
class a {};
```

That's definitely an issue, are you sure the other one is? From some basic testing when I created that error it worked fine, the error is only to prevent human error.

PabstMirror:
```
class b {};
class b {};
```
causes a crash on startup (same as `a` example)
```
ErrorMessage: File x\cba\addons\ai\config.cpp, line 27: .b: Member already defined.
```

BrettMayson:
Alright, Interesting. I remember us talking about this in Discord and I tested it

BrettMayson (commented):
Should change the error code and rename the files to match what is actually going on now

Issue #735, opened 2024-08-07
[Feature Request] Add ability to pass CLI params to hemtt executable for launch presets


DartRuffian:
Allow launch presets to specify hemtt params.

A basic use-case is adding a preset to binarize files and disable filepatching for testing animations quickly. Doing `hemtt launch -b --no-filepatching` is fine, but might be a bit annoying to remember the syntax (i.e. "Is it `-no-filepatching` or `-no-file-patching`" or something) 

Example:
```toml
[hemtt.launch.animations]
hemtt_parameters = [
    "-b",
    "--no-filepatching",
]
```

BrettMayson:
I think just `cli` or `options` would be good, shouldn't be too confusing with context

DartRuffian:
Theoretically it would just be passed to the executable directly, why make it so you can't use the shorthand?
Most people would probably still use the full parameter anyway for readability, but seems like extra logic that doesn't really need to be there

BrettMayson:
It would not be passed to the executable, we're already running

DartRuffian:
Yeah that makes sense, clearly wrote it too early in the morning. Yeah probably only allow fully spelled out parameters for readability

PR #739, opened 2024-08-08
Adds option to add hemtt parameters to launch config

Crowdedlight:
- Adds option to set hemtt cli options in the launch configuration (Issue #735)
- Fixes panic if ``just`` is not present as arg. (Clap update makes it panic now). Do you want this in another separate PR? 

Questions:
1. Currently that function seems to fail either if you set the same parameter in both the config and cli (``hemtt launch nodebug --quick``), or if you use an option that doesn't exist, like ``--quack``. I assume we do not want it to fail and error out if the same option is defined on both CLI and in the config, but "remove duplicates" and carry on? 
2. Wasn't sure to what degree I should make the error message. It could be neat to make it show or print which specific entry in the options was not found, but I might need a pointer on how to build errors of that type and how to detect which one that fails. 
3. wasn't sure on the config entry name. Not sure if just "options" would be better?

Tested:
Tested on my arma3 mod with this launch config, and seemed to work.
```yaml
[hemtt.launch.nodebug]
workshop = [
    "450814997",  # CBA_A3's Workshop ID
    "894678801",  # TFAR Beta
    "1779063631", # Zeus Enhanced 
    "2369477168", # Advanced Developer Tools
    "463939057",  # ACE
    #"1645973522", # Intercept Minimal Dev
    #"1585582292", # Arma Debug Engine
]
mission = "dev_mini.Enoch" # Launch into existing Editor Mission
parameters = [
]
cli_options = [
  "--no-filepatching",
  "--expsqfc",
  "--quick",
]
```

Current Error message if option given is not valid:
billede


bin/src/commands/launch/mod.rs
```diff
@@ -142,6 +143,25 @@ pub fn execute(matches: &ArgMatches) -> Result<Report, Error> {
 
     trace!("launch config: {:?}", launch);
 
+    // extend matches with config args before continuing
+    let Ok(matches) = cli().try_get_matches_from({
```
BrettMayson, line 147: Maybe try `get_matches_from` here too? It looks like it does pretty much the same thing but with error handling

BrettMayson (changes requested):
I think we want to check that every option starts with `--`, to prevent the use of short flags, and to prevent the sort breaking things like
```toml
cli_options = [
  "--optional",
  "some_optional",
]
```

where instead it should be used as
```toml
cli_options = [
  "--optional=some_optional"
]
```

Crowdedlight:
- Changed to ``get_matches_from()`` as suggested for the error handling. 
- Added error message for cases where the cli_options in config doesn't start with ``--``
- Added cli_options to the book under the launch page

Still seems to have some trouble overriding options with default values. Like ``--instances=2`` does not seem to update the default value set in ArgMatches. (I know I would have to update the ``instance_count`` as that is set before the config is loaded, but during testing I tried to update it after loading the config and also inspecting the ``ArgMatches`` afterwards just shows ``instances`` set to ``1`` as default. 

I have not been able to validate that ``--optional=some_optional`` syntax actually works as I don't have any mods with optionals ready, but I do split it into ``["--optional", "some_optional"]`` as clap has it in their documentation. 

I can not get ``--threads`` to work either, neither when set in launch config or just passed on cli. It returns
```
hemtt launch --threads 8
error: unexpected argument '--threads' found
```
This does not happen on main, so my changes have caused ``--threads`` to not be registered anymore. I think it might be because ``let matches = cli().get_matches_from(...)`` is called on ``cli`` which does not have ``threads`` defined, as I assume that is given somewhere else?

BrettMayson:
Would it have made more sense to just add more options to the launch config?

After removing the ones that aren't compatible or are already supported by the launch config, it's really just like quick, no-rap, and no-filepatching that are missing. It would've been simpler to just add those to the launch config

Crowdedlight:
> Would it have made more sense to just add more options to the launch config?
> 
> After removing the ones that aren't compatible or are already supported by the launch config, it's really just like quick, no-rap, and no-filepatching that are missing. It would've been simpler to just add those to the launch config

That is a good point. Makes it more simple!

PR #747, opened 2024-09-01
Lint Configuration

BrettMayson:
- [X] Lints idents no longer include E or W
- [x] Lints can be disabled in TOML
- [X] Lints can have configuration options
- [x] Lints have minimum severity (Can not set a fatal lint to a Warning)
- [x] Convert all Config linting
- [x] Convert all SQF linting
- [x] Automatically generate book documentation from Rust code
- [x] Messages in terminal can link to the book


BrettMayson:
Test coverage isn't great on this, but I have tested everything locally.

Honestly testing could use a refactor as well, would be nice to find a better way to do it and test as many good and bad paths are we can

PR #756, opened 2024-09-02
sqf: fix serilizing empty code

PabstMirror:
maybe alt to #754
I think we still need https://github.com/BrettMayson/HEMTT/pull/748 for extracting vars that are single chars

cba/ace builds using this seem to work fine, but I don't have full confidence


BrettMayson:
Using this branch I don't see any of the issues that #736 fixed for us

PR #757, opened 2024-09-02
utils: fix sqf case util when code doesn't end with whitespace

PabstMirror:
flush buffer if file ends on a char (no whitespace/newline)

e.g. https://github.com/BourbonWarfare/POTATO/blob/master/addons/bodyCleanup/functions/fnc_getPlayerViewCones.sqf#L48
it would delete the last `_viewCones`


BrettMayson (approved):
Ah shoot, good catch

PR #764, opened 2024-09-07
sqf: optimzer - param copies, getvar, vector

PabstMirror:
adds support for 
- `params ["_a", "_b", ["_c", []]];` // makes copy
- `getVariable` / `vectorAdd` 
marginal gains on simpler arrays

on current profiling branch:
10x calls to cba_fnc_getfov: ~0.0378 ms -> 0.0361 ms


PabstMirror:
Using CBA and ACE current branches, here is a `diag_dumpScriptAssembly` dump
one limitation is that it doesn't show the real bytecode for sub code blocks, 
so there are lots of whitespace differences that can be ignored

`--asc`: https://github.com/BrettMayson/HEMTT/blob/1e04e2bfbb7f0aa7f1ce0310b047be316b8014f7/cba_ace.txt
`--expopti`: https://github.com/BrettMayson/HEMTT/blob/af4b0a3dac201d79abb1f67673eb1324e24ba7bd/cba_ace.txt

everything seems to be working as expected
testing in-game also seems to work fine

libs/sqf/src/compiler/optimizer/mod.rs
```diff
@@ -196,34 +207,74 @@ impl Expression {
...
+                    Box::new(Self::ConsumeableArray(array.clone(), range.clone())),
+                    range.clone(),
+                ))
+            }
+        } else {
+            None
+        }
     }
 
-    // Boilerplate for uniary and binary ops
```
BrettMayson, line 226: Unary

BrettMayson (changes requested):

BrettMayson (approved):

PR #767, opened 2024-09-08
Book: Add SQF page for `hemtt utils sqf` command(s)

DartRuffian:
`hemtt utils sqf case` was missing a page in the book.

I don't know what all other setup is required other than creating the markdown file itself.


BrettMayson:
This was intentionally left out, as it can have false positives at the moment and requires manual review

The command and this page need to be "scarier" imo and really push that it needs manual reciews

PR #768, opened 2024-09-08
sqf: Lint for `if ! else`

PabstMirror:
just an idea, probably not worth the effort to gain a tiny bit of speed, also might hurt code flow//readability?

Add suggestion to eliminate `!` when there is an `else`
e.g.
```
        if !(_target in allUnitsUAV) then {
            _target = effectiveCommander _target;
        } else {
            _target = objNull;
        };
```
```
47 │         if !(_target in allUnitsUAV) then {
   │         ^^ if ! else - consider removing not and swapping else
```

### DIFF
--- a/libs/sqf/src/analyze/lints/s10_if_not_else.rs
+++ b/libs/sqf/src/analyze/lints/s10_if_not_else.rs
@@ -1,0 +1,136 @@
+use std::{ops::Range, sync::Arc};
+
+use hemtt_common::config::LintConfig;
+use hemtt_workspace::{lint::{AnyLintRunner, Lint, LintRunner}, reporting::{Code, Codes, Diagnostic, Processed, Severity}};
+
+use crate::{analyze::SqfLintData, BinaryCommand, Expression, UnaryCommand};
+
+crate::lint!(LintS10IfNotElse);
+
+impl Lint<SqfLintData> for LintS10IfNotElse {
+    fn ident(&self) -> &str {
+        "if_not_else"
+    }
+
+    fn sort(&self) -> u32 {
+        80
+    }
+
+    fn description(&self) -> &str {
+        "Checks for unneeded not"
+    }
+
+    fn documentation(&self) -> &str {
+r"### Example
+
+**Incorrect**
+```sqf
+if (!alive player) then { player } else { objNull };
+```
+**Correct**
+```sqf
+if (alive player) then { objNull } else { player };
+```
+
+### Explanation
+
+"
+    }
+
+    fn default_config(&self) -> LintConfig {
+        LintConfig::help()
+    }
+
+    fn runners(&self) -> Vec<Box<dyn AnyLintRunner<SqfLintData>>> {
+        vec![Box::new(Runner)]
+    }
+}
+
+struct Runner;
+impl LintRunner<SqfLintData> for Runner {
+                type Target = crate::Expression;
+    
+    fn run(
+        &self,
+        _project: Option<&hemtt_common::config::ProjectConfig>,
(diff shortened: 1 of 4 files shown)


BrettMayson:
This is one that Cargo suggests, I think it's fine to add as a help lint, people can disable it

PabstMirror:
set to be disabled by default, can be enabled by
```
[lints.sqf]
if_not_else = true
```
could possibly add something like `--wall` to enable all lints via command line?

libs/sqf/src/analyze/lints/s11_if_not_else.rs
```diff
@@ -0,0 +1,134 @@
...
+**Correct**
+```sqf
+if (alive player) then { objNull } else { player };
+```
+`!` can be removed and `else` order swapped
+"
+    }
+
+    fn default_config(&self) -> LintConfig {
+        LintConfig::pedantic() // disabled by default
```
BrettMayson, line 42:
```suggestion
        LintConfig::help()
            .with_enabled(false)
```
I think Help makes more sense for a severity since it is disabled by default. I also prefer being disabled by default in the lint itself, more clear than a different function and a comment

libs/sqf/src/analyze/lints/s11_if_not_else.rs
```diff
@@ -0,0 +1,134 @@
...
+                        return vec![Arc::new(CodeS11IfNot::new(
+                            condition.span(),
+                            processed,
+                            config.severity(),
+                        ))];
+                    }
+                }
+            }
+        }
+        Vec::new()
```
BrettMayson, line 82: I know some of mine need to be revisited, but most of the lints have are flat and use `let ... else `, which imo is more readable for these

libs/sqf/src/analyze/lints/s11_if_not_else.rs
```diff
@@ -0,0 +1,134 @@
...
+    fn severity(&self) -> Severity {
+        self.severity
+    }
+
+    fn message(&self) -> String {
+        "Unneeded not in if".to_string()
+    }
+
+    fn label_message(&self) -> String {
+        "Consider removing and swapping order of else statements".to_string()
```
BrettMayson, line 111:
```suggestion
        "unnecessary `!` operation".to_string()
```

If the label message is too long, it will wrap too often. I would provide more info in help() or note()

libs/sqf/tests/lints/s11_if_not_else/stdout.ansi
BrettMayson, line 1: We'll need a way to test disabled by default lints

BrettMayson (changes requested):

BrettMayson (approved):

libs/sqf/src/analyze/lints/s11_if_not_else.rs
```diff
@@ -0,0 +1,136 @@
...
+    lint::{AnyLintRunner, Lint, LintRunner},
+    reporting::{Code, Codes, Diagnostic, Processed, Severity},
+};
+
+use crate::{analyze::SqfLintData, BinaryCommand, Expression, UnaryCommand};
+
+crate::lint!(LintS10IfNotElse);
+
+impl Lint<SqfLintData> for LintS10IfNotElse {
+    fn ident(&self) -> &str {
```
BrettMayson, line 14:
```suggestion
crate::lint!(LintS11IfNotElse);

impl Lint<SqfLintData> for LintS11IfNotElse {
    fn ident(&self) -> &str {
```

PR #771, opened 2024-09-10
sqf: Fix sqfc byte offset

PabstMirror:
Fixes two problems with sqfc offsets
Shouldn't be any functionality change, just effects arma errors and debugging.

- It would use the macro's start location

```
from sqf:
#define IS_META_SYS(VAR,TYPE) (if (isNil {VAR}) then {false} else {(VAR) isEqualType TYPE})
#define IS_STRING(VAR)   IS_META_SYS(VAR,"STRING")  

before (nonsense because it's using the index of start of macro for both sub blocks):
push [{(if (},{(if (isNil {_pos}) then {fa}]
after:
push [{false},{(_pos) isEqualType "STRING"}]
```

- Unicode would cause problems, as arma seems to expect bytes

e.g. the `#` is in the wrong location
before:
before
after:
after

### DIFF
--- a/libs/sqf/src/compiler/mod.rs
+++ b/libs/sqf/src/compiler/mod.rs
@@ -74,10 +74,7 @@ impl Statements {
                 ctx.add_constant(Constant::String(self.source.clone()))?,
             ))
         } else {
-            let start = processed
-                .mapping(self.span.start)
-                .expect("first statement not in mapping");
-            let offset = start.processed_start().offset() as u32;
+            let offset = processed.get_byte_offset(self.span.start);
             let source = processed.extract(self.span.clone());
             let length = if self.content.is_empty() {
                 0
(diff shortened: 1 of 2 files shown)


PabstMirror:
testing with
```
#define IS_META_SYS(VAR,TYPE) (if (isNil {VAR}) then {false} else {(VAR) isEqualType TYPE})
#define IS_STRING(VAR)   IS_META_SYS(VAR,"STRING")  
#define SOME(A,B) (if (random 1 < 0.5) then { A} else { B })
#define XXX(A,B) (A + B)
#define BAD 5 == "A"

z = IS_STRING(_pos);
```
and
```
#include "macros.inc"

x = "°°°°°°°°";
y = 123456;
if (BAD);
```

sqfc debug output
[before.txt](https://github.com/user-attachments/files/16955146/before.txt)
[after.txt](https://github.com/user-attachments/files/16955148/after.txt)

BrettMayson (approved):

BrettMayson:
Ah nice, good find and fix

PR #776, opened 2024-09-19
sqf: Don't allow trailing commas in arrays

PabstMirror:
Trailing comma in arrays cause script errors (when using non-compiled//filepatching)
```
Error in expression <x = [1,2,]>
Error position: <]>
Error Missing [
```


BrettMayson:
Hm, yeah that's unfortunate

PR #780, opened 2024-09-26
sqf: Inspector

PabstMirror:
- Type checking
- Var Usage (Undefined, Not Private, Shadowing and Unused)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1553,6 +1553,7 @@ dependencies = [
  "hemtt-workspace",
  "linkme",
  "paste",
+ "regex",
  "toml 0.8.19",
  "tracing",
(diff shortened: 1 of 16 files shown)


BrettMayson:
I was also working on something like this, and using the type information from the wiki.

I'm not a fan of it being one big piece that is a single lint

BrettMayson:
Now that we have pedantic, are there pieces of this that can brought into HEMTT with that? Or some parts of this PR that can make it in while others are behind an experimental flag?

BrettMayson:
I think it is time to get this merged

If you think anything should be enabled / moved to pendantic, or put behind an experimental flag let me know

PR #793, opened 2024-10-10
sqf: Add lint for vars that are ALL_CAPS and not a macro

PabstMirror:
Checks for global variables that are ALL_CAPS and may actually be a undefined macro

ref https://github.com/acemod/ACE3/pull/10389
I think default off is a best as some mods use all caps in their code style
e.g. CBA/ACRE `SLX_XEH_MACHINE` `ACRE_SPOKEN_LANGUAGES` `ACRE_IS_SPECTATOR`

### DIFF
--- a/libs/sqf/src/analyze/lints/s17_var_all_caps.rs
+++ b/libs/sqf/src/analyze/lints/s17_var_all_caps.rs
@@ -1,0 +1,128 @@
+use std::{ops::Range, sync::Arc};
+
+use hemtt_common::config::LintConfig;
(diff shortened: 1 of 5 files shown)


BrettMayson:
Default on, but mods can exclude patterns?

Just with ACRE_*,ACE_*, etc .

BrettMayson (approved):
Added a did you mean and context to the diagnostic

image

BrettMayson:
Add label on the source when it is a macro

image

PR #797, opened 2024-10-17
sqf: comparison to static bool

BrettMayson:
Screenshot From 2024-10-16 20-32-37

### DIFF
--- a/libs/sqf/src/analyze/lints/s20_bool_static_comparison.rs
+++ b/libs/sqf/src/analyze/lints/s20_bool_static_comparison.rs
@@ -1,26 +1,23 @@
 use std::{ops::Range, sync::Arc};
 
 use hemtt_common::config::LintConfig;
-use hemtt_workspace::{
-    lint::{AnyLintRunner, Lint, LintRunner},
-    reporting::{Code, Codes, Diagnostic, Processed, Severity},
-};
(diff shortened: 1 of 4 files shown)


libs/sqf/src/analyze/lints/s20_bool_static_comparison.rs
```diff
@@ -0,0 +1,162 @@
...
+        target: &Self::Target,
+        _data: &SqfLintData,
+    ) -> Codes {
+        let Some(processed) = processed else {
+            return Vec::new();
+        };
+        let Expression::BinaryCommand(BinaryCommand::Eq | BinaryCommand::NotEq, lhs, rhs, _) = target else {
+            return Vec::new();
+        };
+
```
PabstMirror, line 66: something like this if you want to match `isNotEqualTo` as well?
```suggestion
        if cmd.as_str() != "==" && cmd.as_str() != "!=" && cmd.as_str() != "isEqualTo" && cmd.as_str() != "isNotEqualTo" {
            return Vec::new();
        }
```

PabstMirror (approved):

libs/sqf/src/analyze/lints/s20_bool_static_comparison.rs
BrettMayson: This creates a bunch of false positives
image

PR #801, opened 2024-10-18
sqf: invalid comparisons

BrettMayson:
There may still be false positives, but I didn't find any with the few repos I have locally
There may be false negatives, but I think that's ok


PabstMirror (approved):
didn't see any hits on my repos

PR #806, opened 2024-10-19
sqf: prevent marker spam

BrettMayson:
image


PabstMirror:
could have something like this (but I doubt anyone does)
```
    while {true} do {
        x setMarkerColor "ColorRed";
        sleep 1;
        x setMarkerColor "ColorBlue";
        sleep 1;
    };
```
and you can just turn the lint off so it's no real problem

PabstMirror (approved):

BrettMayson:
I could wipe the pending calls if it hits a sleep

PabstMirror:
it's fine as is imho

PR #808, opened 2024-10-20
use insta for snapshots

BrettMayson:
I'm curious about your thoughts on this

The tool is nice to review without needing to wipe out the file to have it write the changes like we currently do


PabstMirror:
maybe 2 negatives 
is that any changes to data structure could trigger many things to fail (I think?)
and having to remember to use the cli thing
but overall it seems fine and I do see the benefit

PabstMirror (approved):

PR #812, opened 2024-10-22
fix multibyte character counting in processed

### LINKED ISSUE #749: sqf: unicode problems
- **HEMTT:** main

**Description:**
Unicode in sqf can cause problems for any code bellow it.
Seems like Span//Ranges are sometimes assumed to be by byte instead of by char??

**Steps to reproduce:**
Build with some sqf with unicode:
```
"🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭";
private _workingArms = if (alive player) then { 2 } else { 0 };
```

**Logs:**
```
thread '<unnamed>' panicked at libs\sqf\src\analyze\codes\saa1_if_assign.rs:97:43:
byte index 88 is not a char boundary; it is inside '🌭' (bytes 85..89) of `"🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭";
```

**Additional context:**
I don't think unicode in sqf is that common, 
and I don't think anyone has a real world example of it causing a problem,
so this might be something that can be ignored for now

BrettMayson:
Possibly fix #749 

Fixed the test for format args, might be other places in the code that need fixing, but this will fix position mappings


PabstMirror:
```
"🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭";
private _workingArms = if (alive player) then { 2 } else { 0 };
```
still crashes with
```
Compiling SQF [00:00:00] █░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░      17/504
ERROR panicked at libs\sqf\src\analyze\lints\s05_if_assign.rs:205:43:
byte index 88 is not a char boundary; it is inside '🌭' (bytes 85..89) of `"🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭
🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭🌭";
private _workingArms = if (alive player) then { 2 } else { 0 };
`
```

but it's still an improvement if it's handling single Unicode correctly

BrettMayson:
```rs
let haystack = &processed.as_str()[self.rhs.1.end..];
```

That's going to be the issue within just that lint, I can rewrite it

BrettMayson:
That's fixed now with a test, every usage of `.as_str()[..]` is replaced with `extract()`

PabstMirror (approved):

PR #831, opened 2024-11-09
hooks: ability to call scripts

### LINKED ISSUE #821: Suggestion: run rhai scripts through rhai.
so basically i'd like to be able to run scripts from the hemtt/scripts/ through a rhai command

something like `hemtt().script(scriptname)`

Background:
i currently store all my .rhai files in /scripts/ and copy paste them into the hook folders where i need them.
this way, if i wish to change something, i always have to update multiple files.

If i instead, would be able to run a rhai script through a rhai script, i could keep the main scripts always in the script file and just execute them through a small .rhai script within the hook folder

BrettMayson:
Closes #821


OverlordZorn:
yaay

PR #842, opened 2024-12-03
stringtable: rapify

PabstMirror:
"Binerize" stringtable.xml to stringtable.bin

```
struct XmlbLayout {
    //  4 byte numbers, little end?, nul term strings
    header: Vec<u8>,            // Version
    languages: Vec<u8>,         // Language Count, [Languages]
    offsets: Vec<u8>,           // Offset Count, [Offsets]
    keys: Vec<u8>,              // Key Count, [Keys]
    translations: Vec<Vec<u8>>, // [Translation Count, [Translations], ...] (size may be less than lang count)
```

- All possible languages have to be in the list or the user gets nothing (there is no fallback)
- The number of translation arrays can be less (e.g. just one) so we aren't wasting size
- If a stringtable doesn't have a good "default" language (e.g. just `Dutch/Chinese`) we'll just skip it


libs/stringtable/src/rapify.rs
```diff
@@ -0,0 +1,161 @@
+use crate::{Key, Project};
+use hemtt_workspace::WorkspacePath;
+use tracing::warn;
+
+static ALL_LANGUAGES: [&str; 14] = ["English", "Czech", "French", "Spanish", "Italian", "Polish", "Portuguese", "Russian", "German", "Korean", "Japanese", "Chinese", "Chinesesimp", "Turkish"];
```
dedmen, line 5: Since 2.18 also
Dutch, Finnish, Ukrainian, Swedish, Norwegian, Danish.

See `Arma 3/dta/languageList.bin`

libs/stringtable/src/rapify.rs
```diff
@@ -0,0 +1,161 @@
...
+    for (index, result) in tranlations.into_iter().enumerate() {
+        if let Some(native) = result {
+            all_translations[index].push(native.into());
+        } else if let Some(original) = key.original() {
+            all_translations[index].push(original.into());
+        } else if let Some(english) = key.english() {
+            all_translations[index].push(english.into());
+        } else {
+            // If we don't have some kind of default value to use, we should just not do the conversion
+            return false;
```
dedmen, line 156: BI's binarizer places an empty string here.
But for HEMTT its definitely better if it complains, but it should say where the missing string is? Probably the stringtable linter already does that?

libs/stringtable/src/rapify.rs
BrettMayson, line 5: I would move this to the file where the struct is defined, maybe that file can be redone with codegen at some point to be cleaner

libs/stringtable/src/rapify.rs
```diff
@@ -0,0 +1,161 @@
...
+fn rapify(project: &Project) -> Option<XmlbLayout> {
+    let mut data: XmlbLayout = XmlbLayout::default();
+
+    // Restructure translations: flat for each language
+    let mut all_keys: Vec<String> = Vec::new();
+    let mut all_translations: Vec<Vec<String>> = Vec::new();
+
+    for _language in ALL_LANGUAGES {
+        all_translations.push(Vec::new());
+    }
```
BrettMayson, line 61:
```suggestion
    let mut all_translations: Vec<Vec<String>> = vec![Vec::new(); ALL_LANGUAGES.len()];
```

libs/stringtable/src/rapify.rs
```diff
@@ -0,0 +1,161 @@
...
+}
+fn write_int(buffer: &mut Vec<u8>, input: i32) {
+    buffer.extend(&input.to_le_bytes());
+}
+
+fn rapify(project: &Project) -> Option<XmlbLayout> {
+    let mut data: XmlbLayout = XmlbLayout::default();
+
+    // Restructure translations: flat for each language
+    let mut all_keys: Vec<String> = Vec::new();
```
BrettMayson, line 56: Could be nice to preallocate some space here, but the speed up will be pretty small

libs/stringtable/src/rapify.rs
PabstMirror: https://community.bistudio.com/wiki/Stringtable.xml

not sure if we should add Icelandic, Slovak?  They aren't in languageList.bin
If we don't include a language they get empty strings (there is no fallback)
Don't think it would hurt?

BrettMayson: I would add them, they're already in the big struct

BrettMayson:
Anything outstanding here?

PabstMirror:
I think it's ready

if someone has a stringtable that's just `Dutch/Chinese` we'll just skip it
this should ensure that binned stringtable always work exactly the same as non-binned
IMHO this doesn't need to be a warning

BrettMayson (approved):

PabstMirror:
this breaks escaped symbols like `&apos; &amp; &quot; &lt; &gt; &#...;`
`localize "STR_ACE_Arsenal_statVisionMode_intPrimTi"`= `"Thermal &amp; Primary integrated"`

PR #847, opened 2024-12-18
config: filetype warnings

### LINKED ISSUE #841: Suggestion: Detect if path of class' model property doesnt end in.p3d
I somehow managed to accidentally replace a modelpath's .p3d suffix with .paa.

Maybe thats something that hemtt could watch out for.

Same for .rvmats, .paa, etc.

BrettMayson:
Closes #841 

Sorry for the renames on this PR, should've split that out, noticed it while working on this

image

Originally I had `allow_no_extensions` default to false, but on ACE that resulted in a lot of warnings. Should this be allowed by default?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1568,6 +1568,7 @@ dependencies = [
  "linkme",
  "lsp-types 0.97.0",
  "paste",
+ "toml 0.8.19",
  "vfs",
 ]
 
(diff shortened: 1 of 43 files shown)


PabstMirror (approved):

PabstMirror:
we could always change `allow_no_extensions` default in a followup release so we have time to update projects

PabstMirror:
`soundBegin` / `namesound` seem to be just classnames
 
```
        animationMedic = "AinvPknlMstpSlayW[wpn]Dnon_medicOther";
        animationMedicProne = "AinvPpneMstpSlayW[wpn]Dnon_medicOther";
        animationMedicSelf = "";
        animationMedicSelfProne = "";
```
these are ace specific and reference (sorta) classnames

maybe it should enabled only for specific targets instead of `starts_with`

PR #852, opened 2024-12-19
config: file type check ignores classes

### LINKED ISSUE #851: Config: filetype warnings - False Positives
- **HEMTT:** `1.14.1 stable`
- **Operating System:** Windows

**Description:**
The `allow_no_extension` setting reports lots of false positives since it (seemingly) checks the property names. Below is an incomplete list of false positives. I don't think something like this could ever be 100% compatible with everything without a large amount of hard coding, but in-case it helps.
- `sounds` (in fire mode config): Array of class names to use for sounds
- `soundBegin`: Array of property names and weighted randoms
- `multiSoundHit` Same as `soundBegin`
- `soundBullet`: Sound when a bullet casing hits the ground
- `sound3DProcessingType`: Type of 3D processor to use for sound sets
- `soundShaders`: Array of sound shaders to use in a sound set
- `explosionSoundEffect`: CfgSounds(?) class name
- `nameSound`: For radio protocol, not sure if enum-like value or defined in some class somewhere

**Steps to reproduce:**

1. Enable the `allow_no_extension` setting
2. Run `hemtt check|launch|etc.`

**Expected behavior:**
HEMTT should exclude these (and similar) properties from file extension checking. More reasonably, maybe only specific named properties (e.g. `model`, `picture`, etc.) would only be checked in certain classes? Would need to only be for certain classes for other mods defining their own root classes. For example, [ACE Arsenal Extended](https://github.com/jetelain/AceArsenalExtended/tree/main?tab=readme-ov-file#sample) uses a `model` property that is a class defined elsewhere, not a file path to a `.p3d` file.

**Logs:**
N/A

**Additional context:**
N/A

BrettMayson:
- Assume that values without `\` are classes
- Change `allow_no_extension` to false by default
- Closes #851, probably

### DIFF
--- a/libs/config/src/analyze/lints/c11_file_type.rs
+++ b/libs/config/src/analyze/lints/c11_file_type.rs
@@ -26,11 +26,11 @@ impl Lint<LintData> for LintC11FileType {
     fn documentation(&self) -> &'static str {
 r#"### Configuration
 
-- **allow_no_extension**: Allow properties to not have a file extension, default is `true`.
+- **allow_no_extension**: Allow properties to not have a file extension, default is `false`.
 
 ```toml
 [lints.config.file_type]
-options.allow_no_extension = false
+options.allow_no_extension = true
 ```
 
 ### Example
@@ -114,7 +114,7 @@ impl LintRunner<LintData> for Runner {
         let allow_no_extension = if let Some(toml::Value::Boolean(allow_no_extension)) = config.option("allow_no_extension") {
             *allow_no_extension
         } else {
-            true
+            false
         };
         // Arrays
         if let Value::Array(values) = value {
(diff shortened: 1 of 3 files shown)


PabstMirror (approved):

PabstMirror:
ace has ~800 warnings for the `allow_no_extension` change, but it's also easy enough to turn off

BrettMayson:
They're all legit though? Some evening I could slap extensions on them. This PR should remove all the class ones, are there other false positives

PabstMirror:
I didn't notice any

Another possible issue would be if someone was doing dev builds with wav/png and then did final builds with wss/paa
Such that they want no extensions in the configs
but again they can just turn off the feature

BrettMayson:
Ah, I suppose, I've never done that

PR #859, opened 2025-01-14
sqf: Add lint (S25) for count array comparisons

PabstMirror:
e.g.
```
help[L-S25]: count array comparison
   ┌─ addons/common/functions/fnc_loadPersonLocal.sqf:33:35
   │
33 │         if (count _preferredSeats > 0) then {
   │                                   ^
   │
   = note: use `isNotEqualTo []`
```

edit:
Apparently there is zero (or possibly negative) performance difference using `isEqualTo`
have to look into this more


BrettMayson:
Even if there is no performance difference, I still like this as a readability enabled-by-default lint.

Rust has the same thing for telling you to use `.is_empty()` instead of `.len() == 0`

PabstMirror:
```
a = [1,2,3,4, [5]];
b = [];

20x
count a == 0; 
count b == 0; 
Result:
0.00582418 ms

a isEqualTo []; 
b isEqualTo [];
Result:
0.00563044 ms



a = ["A"];
b = [];
count:
0.00584689 ms
isEqualTo:
0.00560674 ms
```
at least right now on my machine on profiling branch it seems to be a tiny bit faster

BrettMayson:
Is this ready to go? I know the performance difference is very minor, but personally I like it from a code readability perspective, I think its more easily readable at a glance

PR #861, opened 2025-01-16
config: Skip magwell check if no prefix defined

PabstMirror:
e.g. CBA repo
image

this should only be running for magazines that start with `CBA`
So I assume the prefix isn't loaded and is defaulting to "", which matches all mags?


BrettMayson:
This occured on CBA? It has a prefix defined

PabstMirror:
also getting "sound" extension warnings (despite having `options.allow_no_extension = true`)
image

so I guess the real problem is it's not loading the hemtt config at all?
this PR won't fix the core issue so I'll close for now

PR #872, opened 2025-01-18
sqf: Add lint (S26) for short circuit on bool vars

PabstMirror:
`_test1 && _test2` is faster than `_test1 && {_test2}`

But `(!isNil "z") && {z}` is not safe to convert if z is nil

So will probably have to be default off


BrettMayson:
Are there other commands that could break this? We could default on, but skip if either side contains `isNil`

PabstMirror:
`if (GETEGVAR(medical,enabled,false) && {EGVAR(medical_vitals,simulateSpo2)})` from ace will fail

Could maybe do something like:
`hemtt check --pedantic`
to run lints like this that are default off and not something you need to check all the time

BrettMayson:
This PR could be updated to use pedantic and merged?

PR #876, opened 2025-01-22
lints: Check if stringtable entries exist and rework lints (L02)

PabstMirror:
pass buildInfo as the project is being built
currently just stores seen string-table entries but I have other ideas for this as well
unfortunately requires changing function signatures for all lints

- Checks if project stringtable entries exist for both config and sqf
example https://github.com/acemod/ACE3/pull/10684
```
warning[L-C12]: invalid project stringtable entry for config
   ┌─ addons/explosives/TimerDialog.hpp:80:20
   │
80 │             text = CSTRING(Cancel);
   │                    ^^^^^^^^^^^^^^^
   │
   = help: [STR_ace_explosives_Cancel] not in project's stringtables
```


BrettMayson:
Due to #880, we need a different solution for `is_pedantic`, not sure what the best approach is yet, I'd prefer not passing more info around to the lints, and perhaps filtering out the pedantic lints either before / after they run, rather than within the lint and returning an empty vec

libs/workspace/src/lint/mod.rs
```diff
@@ -195,9 +212,12 @@ impl<D> LintManager<D> {
                     }));
                 }
             }
+            if config.enabled() {
+                enabled_lints.push(lint);
```
PabstMirror, line 216: Disabled lints will now be filtered here instead of being checked multiple times below

PabstMirror:
- [ ] ace false positives
```
str_ace_repair_hit - addons/repair/functions/fnc_getHitPointString.sqf:29:16
str_ace_repair_hit - addons/repair/functions/fnc_getHitPointString.sqf:80:13
str_assign_%1 - addons/interaction/functions/fnc_addSquadChildren.sqf:38:37
str_team_%1 - addons/interaction/functions/fnc_joinTeam.sqf:26:37
```

libs/config/src/analyze/lints/collect_stringtables.rs
```diff
@@ -0,0 +1,73 @@
...
+    ) -> Codes {
+        let Some(processed) = processed else {
+            return vec![];
+        };
+        let Value::Str(cstring_data) = target else {
+            return vec![];
+        };
+        let cstring_value = cstring_data.value();
+
+        if cstring_value.to_lowercase().starts_with("str_") || cstring_value.to_lowercase().starts_with("$str_") {
```
PabstMirror, line 60: more self notes

```
x1 = "str_ace_common_cancel";  // "str_ace_common_cancel"
x2 = "$str_ace_common_cancel"; // "$str_ace_common_cancel"
x3 = "$STR_ace_common_cancel"; // "Cancel"
x4 = "$STR_ace_coMMon_cancel"; // "Cancel"

localize "str_aCe_common_canCel" // "Cancel"
localize "$str_aCe_common_canCel" // "Cancel"
```

config needs cased "$STR" to start but then doesn't care about case
sqf doesn't care about case or if starting with a $
this might be newish - otherwise I don't know why we have CSTRING and LSTRING macros?

jokoho48: After the wiki $ in localize was introduced in 2.04

PabstMirror:
Ace repair hit still is a false positive - ref https://github.com/acemod/ACE3/blob/master/addons/repair/functions/fnc_getHitPointString.sqf#L29

but I've added config options so it can be ignored
```
[stringtables.usage]
options.ignore = ["str_ace_repair_hit"]
options.ignore_unused = true
```

libs/stringtable/src/lib.rs
```diff
@@ -43,51 +46,62 @@ static ALL_LANGUAGES: [&str; 25] = [
...
-    #[serde(rename = "@name")]
-    name: String,
-    #[serde(rename = "Package")]
-    packages: Vec<Package>,
-
-    #[serde(skip)]
-    meta_comments: Vec<(String, String, Option<String>)>,
+    inner: InnerProject,
+    path: WorkspacePath,
+    keys: HashMap<String, Vec<Position>>,
```
PabstMirror, line 53: hashmaps have random order
so I think this is why the `insta` tests are failing?

PabstMirror: hoping 0de59759022443bcd0b0a3d3690a265d781423f3 makes sense
could also zero out `keys` as it really isn't the focus of the sort testing

BrettMayson (approved):
I think we're all good here, maybe some minor future tweaks but I'm good for you to merge if you're happy too

libs/stringtable/src/analyze/lints/l02_usage/mod.rs
```diff
@@ -0,0 +1,255 @@
...
+                pos.path().as_str().trim_start_matches('/'),
+                pos.start().1 .0,
+                pos.start().1 .1
+            )
+            .expect("Failed to write to file");
+        }
+        if !ignore {
+            codes.push(Arc::new(CodeStringtableMissingFile::new(
+                missing.len() as u64,
+                Severity::Error,
```
PabstMirror, line 216:
```suggestion
                Severity::Warning,
```
ace does have a false positive on this so it might be better as a warning for a short time?

BrettMayson: Ah, yeah should be warning overall imo, doesn't need to be an error

PR #880, opened 2025-01-25
Stringtables - Use addon build_data

BrettMayson:
Changes the implementation of #876 

1. The Config and SQF modules always run before the Stringtables module
2. Config and SQF modules store their stringtable usage in `addon.build_data()`
3. A stringtable lint `L02` goes through the usage and stringtables, finding missing, unused, duplicates, etc.

This prevents the need for the buildinfo to be passed around everywhere, using instead the existing LintData that is passed to all lints. This will also provide stringtable usage data in a clean to the ls / possible future utils, which could be useful

- Also changes the lint that used `is_release` to have `always` and `release` options instead

- Needs something for `is_pedantic` still, not sure how I best want to handle this yet, but want to get these changes up

### DIFF
--- a/arma/src/lib.rs
+++ b/arma/src/lib.rs
@@ -30,7 +30,7 @@ fn init() -> Extension {
                 && len_buf != [255u8; 4]
             {
                 let len = u32::from_le_bytes(len_buf);
-                println!("Receiving: {}", len);
+                println!("Receiving: {len}");
                 let mut buf = vec![0u8; len as usize];
                 socket.read_exact(&mut buf).unwrap();
                 let buf = String::from_utf8(buf).unwrap();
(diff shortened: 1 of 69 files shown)


BrettMayson:
image

Example of finding unused

PabstMirror:
sorry about the `build_info_catastrophe`
I really wasn't sure the best way to move that data around

if the stringtable location parsing is trouble, it might not be needed
Just knowing the name of the duplicate should be good enough

BrettMayson:
Going to merge this into the other PR, I'm working on an error generation library that handles graphemes correctly, so we can use it on stringtables with all sorts of characters, but that'll require a new PR to rework the internals of HEMTT a bit anyway. I think after this is merged the original PR should be good to go to? unless there's anything else that should be done now for it

PR #891, opened 2025-02-20
sqf: disable _this call lint

BrettMayson:
The performance difference is minimal, and people seem to prefer with `_this`

Perhaps this can be done in the optimizer instead?


PabstMirror:
https://github.com/BrettMayson/HEMTT/blob/main/libs/sqf/src/compiler/optimizer/mod.rs#L147-L153

PR #892, opened 2025-02-20
sqfc: use #line

BrettMayson:
Using #line as per the discussion on Discord


BrettMayson:
@PabstMirror I've tested this with ACE, but only on single player. I have found no issues, would you be able to run a test as well, confirm everything looks good

PabstMirror:
my group's addon saw a significant time increase for check/release
from 4.5 to 33 seconds

most of it locked in `Rapifying Configs 89/90`
specifc one being https://github.com/BourbonWarfare/POTATO/tree/master/addons/miscFixes/patchHAFMHelis
 which does have some massive mfd includes, total preprocessed is ~20k lines

BrettMayson:
I can look into performance in follow up PRs, did you find any issues with the output of the builds?

PabstMirror:
everything built seemed ok

BrettMayson:
Well this PR is the way we want to move, I'm going to merge it and will do some performance profiling sometime in the next week. I don't know if we want to do a HEMTT RC with this (an ACE RC built with said HEMTT RC would be really great)

PabstMirror:
the funny thing is I don't really see any change to preprocess/config that would cause the slowdown I'm seeing
and I know ace/cba have some big configs as well

example.sqf
```
params ["_x"];

if (_x isEqualType "A") then {
    _x + 1;
};
#include "example.inc.sqf"
systemChat "done";
```
inc:
```
if (_x isEqualType 1) then {
    _x + "A";
};
```

old hemtt:
```
Error in expression <
};
if (_x isEqualType 1) then {
    _x + "A";
};
systemChat "done";>
  Error position: <+ "A";
};
systemChat "done";>
  Error Generic error in expression
File /x/cba/addons/settings/example.inc.sqf..., line 2
```

this PR:
```
Error in expression <f"
if (_x isEqualType 1) then {
    _x + "A";
};
#line 7 "\addons\settings\examp>
  Error position: < "A";
};
#line 7 "\addons\settings\examp>
  Error Generic error in expression
File /x/cba/addons/settings/example.inc.sqf..., line 2
```

asc:
```
Error in expression <qf"
if (_x isEqualType 1) then {
    _x + "A";
};

#line 6 "/x/cba/addons/settin>
  Error position: <+ "A";
};

#line 6 "/x/cba/addons/settin>
  Error Generic error in expression
File //x/cba/addons/settings/example.inc.sqf..., line 2
```

compile preprocessFileLineNumbers:
```
Error in expression <nc.sqf"
if (_x isEqualType 1) then {
_x + "A";
};
#line 6 "f:\dev\localmissions\>
  Error position: <+ "A";
};
#line 6 "f:\dev\localmissions\>
  Error Generic error in expression
File f:\dev\localmissions\missions\testinguavzeus.stratis\asc\example.inc.sqf..., line 2
```

`#line 7 "\addons\settings\examp>`
ASC and in-game both think it should be line 6? but I'm not sure
but I'm pretty sure it should be prefixed with the full in-game path not the relative project path
so `\x\cba\...`

libs/workspace/src/reporting/processed.rs
```diff
@@ -173,6 +189,65 @@ fn append_output(
...
+
+        let Some(map) = processed.mapping(cursor_offset + 1) else {
+            panic!("mapping not found for offset {cursor_offset}");
+        };
+        let pending_line = comitted_line + pending_empty;
+        let linenum = map.original().start().line();
+        let file = map
+            .original()
+            .path()
+            .as_str()
```
PabstMirror, line 219:
```suggestion
            .as_virtual_str()
```
I think?

BrettMayson:
> I don't really see any change to preprocess/config that would cause the slowdown

There were also some changes for the HLS, there was a regression in performance but I fixed that up later, perhaps the slowdown is coming from an edge case there rather than this PR?

I will do some performance profiling, HEMTT's due for it again anyway. I wish it were easy to setup CI performance benchmarks, so we could see that in PRs easily

PabstMirror:
retested my project
main is 5 seconds
this pr is 42
so it must be something in here

PabstMirror (approved):

BrettMayson:
Hm, odd, not sure what it could be, but I'm sure I'll find it with a profiler

PR #893, opened 2025-02-21
sqf: Support reverse chained math ops

### LINKED ISSUE #890: Optimizer - Operation order with a variable
This gets optimized to a single multiplication command

```sqf
private _seconds = 24 * 60 * 60 * _days;
```

This does not

```sqf
private _seconds = _days * 24 * 60 * 60;
```

PabstMirror:
Fix #890

transforms:
(X / 2) * 5
X / (2 / 5)
X / 0.4


PabstMirror:
could do this on strings as well for addition, but I saw ~2 uses in all of cba and ace so it's just not worth it imho

BrettMayson (approved):

PR #894, opened 2025-02-22
stringtables: Warn on unsupported stringtable formats, ignore Stringtable.xml case

PabstMirror:


bin/src/modules/stringtables.rs
```diff
@@ -53,7 +53,19 @@ impl Module for Stringtables {
                 .walk_dir()
                 .expect("vfs issue")
                 .into_iter()
-                .filter(|p| p.filename() == "stringtable.xml")
+                .filter(|p| {
+                    if p.filename() == "stringtable.xml" {
```
BrettMayson, line 57: Can we do to_lowercase here?

BrettMayson (approved):

PR #899, opened 2025-02-27
preprocessor: mapping iter

BrettMayson:
Uses an iterator instead of collecting the mappings into a vec, needs to filter less mappings and avoids allocating vecs

`hemtt dev --just miscFixes -t 1` on POTATO went from 46 to 28 on my machine. This is a good change, but still far from a solution


BrettMayson:
The performance impact after the regression was fixed is minimal, and I think the code is slightly more readable as a vec, so I think I'll leave this out, but this is a possible avenue of performance tweaks in the future, on this and other vecs

PR #905, opened 2025-03-01
sqf: fix sqfc code spans

BrettMayson:
off by one

also renamed total to total_chars for clarity


PabstMirror:
somehow this PR has caused a performance regression
my mod has gone from 4.5sec to 14.2, again mostly on 90/91 configs

PR #909, opened 2025-03-05
config: Add lint C12 for quoted math that could be run at build-time

PabstMirror:
Adds pedantic lint for quoted math statements that hemtt could handle

```
help[L-C12]: Math could be unquoted
  ┌─ addons/main/config.cpp:5:28
  │
5 │         requiredVersion = "2 + 0.18";
  │                            ^^^^^^^^ reducible to: 2.18
  │
  = note: Could remove quotes to allow evaluation at build-time
```

### DIFF
--- a/libs/config/src/analyze/lints/c12_math_could_be_unquoted.rs
+++ b/libs/config/src/analyze/lints/c12_math_could_be_unquoted.rs
@@ -1,0 +1,174 @@
+use std::{ops::Range, sync::Arc};
+
+use hemtt_common::config::{LintConfig, LintEnabled, ProjectConfig};
+use hemtt_workspace::{
+    lint::{AnyLintRunner, Lint, LintRunner},
+    reporting::{Code, Diagnostic, Processed, Severity},
+};
+
+use crate::{analyze::LintData, Item, Number, Value};
+
+crate::analyze::lint!(LintC12MathCouldBeUnquoted);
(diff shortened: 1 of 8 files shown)


libs/config/src/analyze/lints/c12_math_could_be_unquoted.rs
```diff
@@ -0,0 +1,174 @@
...
+```
+
+### Explanation
+Quoted math statements will have to be evaulated in-game
+"
+    }
+
+    fn default_config(&self) -> LintConfig {
+        // false-positives are possible
+        LintConfig::help().with_enabled(LintEnabled::Pedantic)
```
BrettMayson, line 46: You've seen false positives from this? I would think it should be possible to avoid them

PabstMirror: just possible in theory as some kind of formatted text
like `text = "0000-9999" // input range`

BrettMayson: Ah, true

BrettMayson: Personally I think this would be fine to enable by default, I think the odds of encountering text like that is low enough that the benefit is worth it

PR #921, opened 2025-03-11
sqf: lint S27 for select with count

BrettMayson:


PabstMirror (approved):

BrettMayson:
Yeah that'd really complicate it to try and check for that, probably easier to pressure BI to make it work lol

PR #924, opened 2025-03-12
config: Have c12 ignore configs that are usually strings

PabstMirror:
- `text = "0-9"; // ignored because name`
blocklist "text", "name", "displayname" should reduce any possible false positives

- ` initSpeed = "300"; // forced because name`
not sure about this part or what entries should be, just an option

### DIFF
--- a/libs/config/src/analyze/lints/c12_math_could_be_unquoted.rs
+++ b/libs/config/src/analyze/lints/c12_math_could_be_unquoted.rs
@@ -94,7 +94,7 @@ impl LintRunner<LintData> for Runner {
         let check_if_equation = !match config.option("forced") {
             Some(toml::Value::Boolean(forced)) => *forced,
             Some(toml::Value::Array(forced)) => forced.iter().map(|v| v.as_str().expect("forced items must be strings").to_lowercase()).any(|x| x == name.as_str()),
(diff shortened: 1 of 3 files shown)


PabstMirror:
```
a = 1;
b = "1";
c = "1+0";
getNumber (configFile >> "a"); x20 
0.00561725 ms 
getNumber (configFile >> "b"); x20 
0.00701223 ms
getNumber (configFile >> "c"); x20 
0.0367034 ms
```

at least speed wise stringed-numbers aren't a big difference

but this idea came from:

Dedmen — Today at 9:16 AM
> These simple maths are actually somewhat a problem.
Scripting is not multithreading safe, and these execute scripts, and since recently, we load things like materials in MT.
There is some material in vanilla that does useWorldEnvMap = "true" which triggers script execution in MT and can crash the game...
I optimized away the "true"/"false" cases to fix that problem for materials. The world would be much better if people weren't trying to do that stuff.
Hopefully now that hemtt detects it, everyone fixes their issues.
I assume hemtt already detects = "true" in configs?

so maybe configs like useWorldEnvMap, specularPower, ambient, diffuse

PabstMirror:
I think this is an ok start for now

PR #929, opened 2025-03-13
wss: initial support

BrettMayson:
- [x] WSS Files
- [x] Reading
- [x] Writing
- [x] No compression
- [x] Byte compression
- [x] Nibble compression
- [x] WAV Files
- [x] OGG Files
- [x] MP3 Files
- [ ] LIP Files (?)
- [x] Inspect util
- [x] Conversion util
- [x] Extension Convert
- [x] Extension Preview

It'd be nice for HEMTT to have a util to play audio, but that seems a bit much imo. I think previewing in the extension is good enough

LIP files, I have not found a spec on, but since it is already an imperfect solution, I think as long as HEMTT could create something decent that'd be fine.


libs/wss/src/lib.rs
```diff
@@ -0,0 +1,134 @@
...
+    /// # Errors
+    /// [`std::io::Error`] if an IO error occurs
+    /// [`Error::UnsupportedFileType`] if the file is not a WSS file
+    /// [`Error::InvalidCompressionValue`] if the compression value is invalid
+    pub fn read<I: Read + Seek>(input: &mut I) -> Result<Self, Error> {
+        let mut reader = BufReader::new(input);
+
+        let mut buffer = [0; 4];
+        reader.read_exact(&mut buffer)?;
+        if &buffer != b"WSS0" {
```
dedmen, line 36: missmatch means the WSS header (8 bytes) is not present, the file does not use compression (compression=0) and the first byte is the format of the wave header

libs/wss/src/lib.rs
```diff
@@ -0,0 +1,134 @@
...
+            return Err(Error::UnsupportedFileType(
+                String::from_utf8_lossy(&buffer).to_string(),
+            ));
+        }
+
+        let compression = reader.read_u32::<LittleEndian>()?;
```
dedmen, line 42: compression is actually a single byte, and 3 bytes padding.
There is no guarantee for the padding to be zeroed

libs/wss/src/wav.rs
```diff
@@ -0,0 +1,72 @@
...
+
+    /// Create a new WSS file from a WAV file.
+    ///
+    /// # Errors
+    /// [`Error::Wav`] if an error occurs while reading the WAV file.
+    pub fn from_wav<R: Read>(wav: R) -> Result<Self, Error> {
+        Self::from_wav_with_compression(wav, &Compression::default())
+    }
+
+    /// Create a new WSS file from a WAV file with a specific compression type.
```
dedmen, line 40: Compression types have preconditions, I can't see checks for them?
bits per sample must be 16, channels must be 1 or 2.

libs/wss/src/wav.rs
```diff
@@ -0,0 +1,72 @@
...
+    /// # Errors
+    /// [`Error::Wav`] if an error occurs while reading the WAV file.
+    pub fn from_wav_with_compression<R: Read>(
+        wav: R,
+        compression: &Compression,
+    ) -> Result<Self, Error> {
+        let mut reader = hound::WavReader::new(wav)?;
+        let channels = reader.spec().channels;
+        let sample_rate = reader.spec().sample_rate;
+        let bits_per_sample = reader.spec().bits_per_sample;
```
dedmen, line 51: Compression types have preconditions, I can't see checks for them?
bits per sample must be 16, channels must be 1 or 2.

libs/wss/src/compression/byte.rs
```diff
@@ -0,0 +1,137 @@
+#[allow(clippy::excessive_precision)]
+const LN_10: f64 = 2.302_585_092_994_045_684_084_0;
+#[allow(clippy::excessive_precision)]
+const LOG2: f64 = 1.442_695_040_888_963_407_0;
+
+#[allow(clippy::suboptimal_flops)]
+pub fn decompress(data: &[u8]) -> Vec<i16> {
+    let magic_number: f64 = (LN_10 * LOG2) / 28.125_740_425_151_72;
+    let mut output_data = Vec::with_capacity(data.len());
+    let mut last_val: i16 = 0;
```
dedmen, line 10: Again this is the mono only. Stereo data has two delta's.
Also delta8 compression is very similar to delta4.
Except that the lookup table has 255 values, where delta4 has 15.

Why is this code so much different 🤔

libs/wss/src/compression/nibble.rs
```diff
@@ -0,0 +1,156 @@
...
+const PCM_INDEX: [i16; 16] = [
+    -8192, -4096, -2048, -1024, -512, -256, -64, 0, 64, 256, 512, 1024, 2048, 4096, 8192,
+    0, // Last value is padding for symmetry
+];
+
+/// Decompress nibble-compressed data back to PCM samples.
+#[allow(clippy::cast_possible_truncation)]
+pub fn decompress(data: &[u8]) -> Vec<i16> {
+    let mut output = Vec::with_capacity(data.len() * 2);
+    let mut delta: i32 = 0;
```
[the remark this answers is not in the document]
BrettMayson: This is based on the C++ example from the biwiki, and seemed to work fine on the ADR 97 files, which have two channels

libs/wss/src/compression/byte.rs
dedmen: A good test might be
A3/Sounds_F_Enoch/Assets/Environment/Backgrounds/Meadow/Meadow_Bg_Day_Quiet
Our encoder used to mess up on that until last update.

> Bis optionally use Delta8 (byte) or Delta4 (nibble) 'Lossy' Compression for mono pcm data only.
> The facility exists to handle stereo compression but it isn't employed. 

That may have been right when this was written 15 years ago. it isn't anymore.

dedmen:
Its confusing to me that the compression types are named Byte and Nibble. Which missmatches the naming in the official tools (delta4, delta8)
```
Usage:  WaveToSSS <options> <source> [<destination>]
Options:-d8 8b deltapack
        -d4 4b deltapack (voice)
```

BrettMayson:
Just what it was called in the wss biwiki page, but yeah I can change it

PR #940, opened 2025-03-17
config: Fix C12 case sensitivity on ignore option

PabstMirror:
ref https://discord.com/channels/976165959041679380/1039189045655392389/1351199509094600846


BrettMayson (approved):

BrettMayson:
Should also update the documentation function

I might make it a string instead of a static string slice so we could move the arrays out and use format!

PR #941, opened 2025-03-18
extended update message with instructions

R3voA3:


BrettMayson:
I don't think this should be needed, it'll already tell you to update via Winget if that's how you have it installed

PR #979, opened 2025-05-25
rapify: support more extensions

### LINKED ISSUE #975: Feature request - Rapify all rapifiable files (.sqm, .fsm, .bikb, .bisurf)
Currently HEMTT rapifies only `config.cpp`, `*.ext` and `*.rvmat` files. I suggest it should rapify all rapifiable files, which also include `*.sqm`, `*.fsm`, `*.bikb` and maybe `*.bisurf` files. (I haven't actually seen a `.bisurf` file but I heard they use the same config class syntax.)

**Advantages:**

- The rapifier would syntax check those files for missing commas etc.
- The engine would probably load them faster.

**Disadvantages:**

I'm not an expert on Arma file formats so I don't know why someone would want non-rapified files in their build. Maybe it has some benefits in debugging. At least non-rapified files are human readable.

**Has this been done elsewhere?**

- Arma 3 by default rapifies all `mission.sqm` files saved in the editor.
- Mikero's MakePbo tool promises (at least 9 years ago) to rapify all these seven formats according to this old documentation
https://github.com/Anrop/docker-mikero-tools/blob/e862157ee6f2a34f711096d4838c59bbe42fa78c/doc/depbo-tools-0.5.50/MakePbo.ReadMe.htm#L75

BrettMayson:
Closes #975

### DIFF
--- a/bin/src/modules/rapifier.rs
+++ b/bin/src/modules/rapifier.rs
@@ -222,5 +222,5 @@ pub fn can_rapify(path: &str) -> bool {
             path.trim_start_matches('/')
         );
     }
-    ["cpp", "rvmat", "ext"].contains(&ext)
+    ["cpp", "rvmat", "ext", "sqm", "fsm", "bikb", "bisurf"].contains(&ext)
 }


PabstMirror:
ace gets expected errors on .bisurf
```
impact = Hit_Glass;
soundHit = glass;
```
which are easy fixes

one thing that might be a problem is if someone puts pre-binerized missions.sqm into an addon (like what the editor will save as)

e.g. a sqm that is like
```
 raP       Rt  	version 6    EditorData
```
will cause
```
ERROR Failed to execute command:
Preprocessor error: Workspace Error: VfsError: Could not read path for '/addons/ai/mission.sqm': IO error: stream did not contain valid UTF-8
```

BrettMayson:
Ah good point, didn't think about already rapified files

PR #995, opened 2025-05-27
config: fix processing speed

BrettMayson:
@PabstMirror can you test this with your mod that had the one addon that was processing very slow

This fixes the speed issue, but I can't remember if there was another issue that was encountered instead

### DIFF
--- a/libs/workspace/src/reporting/processed.rs
+++ b/libs/workspace/src/reporting/processed.rs
@@ -125,7 +125,7 @@ fn append_token(
         });
         processed
             .mappings_newlines
-            .push((processed.total_chars, processed.mappings.len()));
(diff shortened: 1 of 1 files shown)


BrettMayson:
Can you confirm that the SQFC is still good? iirc this issue came up while we were making changes to sqfc. I think that's why this wasn't fixed before

BrettMayson:
The SQFC tests pass on all of the different ways I modified this :|

If you could test the latest version, that'd be greatly appreciated, if we find any SQFC problems we definitely to add more edge cases

BrettMayson:
I added a debug to compare SQFC between builds, and yeah this fails on ACE

PR #996, opened 2025-05-28
workspace: use interval tree for mappings

BrettMayson:
Not as fast as #995, but with the massive benefit of working

The SQFC files match main testing against current master ace


PabstMirror:
now getting 4.6 sec on that project (from 15.6 -> 4.8 on 995)

PabstMirror (approved):

BrettMayson:
Interesting, for me it was consistently a bit slower, 58s to 3.1, where 995 was 1.7

PR #1007, opened 2025-06-18
fnl - ignore license.txt

PabstMirror:
Shouldn't really matter but I'd rather not modify licenses (if they need to be included untouched)

### DIFF
--- a/bin/src/modules/fnl.rs
+++ b/bin/src/modules/fnl.rs
@@ -35,7 +35,7 @@ impl Module for FineNewLineCheck {
                 .filter(|e| {
                     e.path()
                         .file_name()
-                        .is_some_and(|f| f.to_ascii_lowercase() != "license.txt")
+                        .is_some_and(|f| !f.eq_ignore_ascii_case("license.txt"))
                 })
                 .map(|e| e.path().to_path_buf())
                 .collect::<Vec<_>>()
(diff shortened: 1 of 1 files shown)


bin/src/modules/fnl.rs
```diff
@@ -32,6 +32,11 @@ impl Module for FineNewLineCheck {
                         .extension()
                         .is_some_and(|e| TEXT_EXTENSIONS.contains(&e.to_str().unwrap_or_default()))
                 })
+                .filter(|e| {
+                    e.path()
+                        .file_name()
+                        .is_some_and(|f| f.to_ascii_lowercase() != "license.txt")
```
PabstMirror:
```suggestion
                        .is_some_and(|f| !f.eq_ignore_ascii_case("license.txt"))
```

BrettMayson:
This seems incredibly silly to add code for that runs every time HEMTT does, when I can not see this ever mattering at all

This also won't prevent `hemtt utils fnl` from adding one anyway

Also this makes me think we could include more extensions, like md, into this

BrettMayson:
Yeah, I'm going to say no to a merge. This just doesn't matter, someone copy pasting the text could change this, or the encoding, or .md vs .txt vs no extension, it just doesn't matter and isn't worth the code in my opinion

PR #1008, opened 2025-06-19
config: Check if parent exists in scope

### LINKED ISSUE #956: config: Check if parent exists in scope
- **HEMTT:** hemtt 1.15.1
- **Operating System:** [e.g. Windows, macOS, Linux]

**Description:**
Doesn't check if parent exists in scope

**Steps to reproduce:**
```
class CfgVehicles {    
    class Logic;
    class Module_F: Logic {
        class AttributesBase {
            class Combo;
        };
    };
    class test: Module_F {
        scope = 2;
        displayName = "test";
        class Attributes { // should inherit from AttributesBase
            class Side: Combo {
                displayName = "example";
            };
        };
    };
};
```

**Expected behavior:**
Should report as error
When launching arma loads ok and just reports
`Cannot find base Combo for z\ace\addons\ai\config.bin/CfgVehicles/test/Attributes/Side/`

If launching same config with no-rap it errors out
```
ErrorMessage: File z\ace\addons\ai\config.cpp, line 28: /CfgVehicles/test/Attributes.Side: Undefined base class 'Combo'
```

**Logs:**
If applicable, add logs to help explain your problem.

**Additional context:**

PabstMirror:
Fix #956

### DIFF
--- a/libs/config/tests/lints/c05_parent_case.hpp
+++ b/libs/config/tests/lints/c05_parent_case.hpp
@@ -3,12 +3,12 @@ class local: Imported {
     value = 1;
 };
 
-class A {
-    class B {};
-    class C {};
-};
+class B {};
+class C {};
 
+class A {
 class b: B {};
 
 // Will be ignored, because the parent case can match the child case
 class c: c {};
+};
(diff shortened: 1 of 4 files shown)


libs/config/tests/lints/c05_parent_case.hpp
```diff
@@ -3,12 +3,12 @@ class local: Imported {
...
-    class C {};
-};
+class B {};
+class C {};
 
+class A {
 class b: B {};
 
 // Will be ignored, because the parent case can match the child case
 class c: c {};
```
PabstMirror:
```suggestion
    class b: B {};

    // Will be ignored, because the parent case can match the child case
    class c: c {};
```

PR #1009, opened 2025-06-19
config: C14 Check unused externs

PabstMirror:
might be a smell that something is wrong with configs
unless there is a legit reason to extern things that are not used?


libs/config/src/analyze/lints/c14_unused_external.rs
```diff
@@ -0,0 +1,243 @@
...
+class x;
+```
+
+**Correct**
+```hpp
+```
+"
+    }
+    fn default_config(&self) -> LintConfig {
+        LintConfig::help().with_enabled(hemtt_common::config::LintEnabled::Pedantic)
```
BrettMayson: Is there a good reason for pedantic?

Would this be prone to false positives or genuine reasons to exist?

PabstMirror: They don't really create any problems and I don't want to force people to change good working configs
vehicle turret configs especially may require many lines changed

but I still think it can be a useful optional tool to show configs that may be inheriting from the wrong level
or if people really want to cleanup as much as possible

BrettMayson: I'd kind of lean to opt out then personally, no one reads the options, and if this is opt in, it'll get used by nearly no one

BrettMayson: Perhaps a thing where if there are more than x diags from this lint, we suppress them by default and tell the user to either turn it off or explicitly turn it on?

PR #1016, opened 2025-06-24
config: enable c14 by default

BrettMayson:
image

Similar to the unused stringtables

If we go this way instead, do we want to add paths to ignore?

I'm also ok reverting the -Lc14 thing, to be honest I started that, then remembered txt files exist, and did both, but I don't think we *need* it. it changes the lints to pass RuntimeArguments rather than just the bool for pedantic, and those arguments are passed down to the lints, means every lint needs to be modified to add that, but we can reuse that in the future to pass more runtime info down to the lints if needed

Only going to bother with tests if you're ok going this way

### DIFF
--- a/bin/src/commands/check.rs
+++ b/bin/src/commands/check.rs
@@ -27,6 +27,9 @@ pub struct CheckArgs {
     #[arg(long, short = 'p', action = clap::ArgAction::SetTrue)]
     /// Run all lints that are disabled by default (but not explicitly disabled via project config)
     pedantic: bool,
+    #[arg(long, short = 'L', action = clap::ArgAction::Append)]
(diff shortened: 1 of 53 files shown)


bin/src/commands/check.rs
```diff
@@ -27,6 +27,9 @@ pub struct CheckArgs {
     #[arg(long, short = 'p', action = clap::ArgAction::SetTrue)]
     /// Run all lints that are disabled by default (but not explicitly disabled via project config)
     pedantic: bool,
+    #[arg(long, short = 'L', action = clap::ArgAction::Append)]
+    /// Run all lints that are disabled by default (but not explicitly disabled via project config)
```
PabstMirror:
```suggestion
    /// Explicit Lints
```

PabstMirror (approved):

PR #1021, opened 2025-06-29
config: add flag to output derap'd configs as json

TheCandianVendingMachine:
`Array::Item::*` is now pub, this matches the accessibility given by every other config model

I don't know if this is where the json serialization should live. I didn't put it in the library crate to avoid adding un-needed dependencies.

json serialization uses recursion for implementation simplicity


BrettMayson:
I'd prefer that serialize be implemented on the config types directly within the config library, rather than as a function in the binary utils

If can make that change if you're unfamiliar

TheCandianVendingMachine (commented):
outputted json sucks, give me a bit

TheCandianVendingMachine:
writing a deserializer implementation would be hell, and adds nothing to this pr. I'd prefer to not at the moment

otherwise this produces a json dump that is fairly good

PR #1028, opened 2025-07-14
hls: completions

BrettMayson:


libs/config/tests/snapshots/lints__config_error_c14_unused_external.snap
```diff
@@ -21,3 +21,10 @@ expression: lint(stringify! (c14_unused_external)).0
    [0m[36m│[0m
 [0m[36m18[0m [0m[36m│[0m     class [0m[33mDefault[0m; // Not Used
    [0m[36m│[0m           [0m[33m^^^^^^^[0m [0m[33mnever used[0m
+
+
+[0m[1m[38;5;11mwarning[L-C14][0m[1m: external class beta is never used[0m
```
PabstMirror: I don't like seeing duplicate warnings for the same line

BrettMayson: Yeah I'm not sure yet why this popped up on this PR

PR #1039, opened 2025-07-24
preprocessor: add array support for macro parameters

tyen901:
Arrays containing commas can now be used as macro arguments without being split. The preprocessor now tracks brace depth in addition to parentheses depth when parsing macro arguments, ensuring that commas inside braces are not treated as argument separators.

Changes:
- Add brace parsing to pest grammar and parser
- Track brace depth in call_read_args()
- Add is_left_brace/is_right_brace methods to Symbol
- Update argument splitting logic to respect brace boundaries
- Add comprehensive tests for array macro functionality

### DIFF
--- a/libs/preprocessor/tests/bootstrap/array_support/source.hpp
+++ b/libs/preprocessor/tests/bootstrap/array_support/source.hpp
@@ -1,0 +1,5 @@
+#define VERSION_ARRAY(major, minor, patch) {major, minor, patch}
+#define ARRAY_MACRO(name, arr) class name { version[] = arr; };
+
+version = VERSION_ARRAY(1, 2, 3);
+ARRAY_MACRO(test_mod, {3, 15, 7})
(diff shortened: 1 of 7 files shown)


libs/preprocessor/tests/bootstrap/array_support/source.hpp
```diff
@@ -0,0 +1,5 @@
+#define VERSION_ARRAY(major, minor, patch) {major, minor, patch}
+#define ARRAY_MACRO(name, arr) class name { version[] = arr; };
+
+version = VERSION_ARRAY(1, 2, 3);
+ARRAY_MACRO(test_mod, {3, 15, 7})
```
PabstMirror, line 5: unfortunately arma still can't read these configs
```
File z\ace\addons\cargo\config.cpp, line 30: '/test.version': Missing ';' prior '}'
Error context )
ErrorMessage: File z\ace\addons\cargo\config.cpp, line 31: '.7': '}' encountered instead of '='
```

this will cause problems if doing `--no-rap` for filepatching/`diag_mergeConfigFile`

it's a similar problem to allowing trailing commas
one possible solution is to add a option to allow these types of hemtt-only configs?

tyen901:
That's a shame. So it sounds like this idea is dead in the water?

BrettMayson:
Closing as I don't think this can be added

PR #1087, opened 2025-10-08
Inspector - Test all wiki examples

PabstMirror:
tests every single wiki example for bad args
more of a tool to improve the wiki

### DIFF
--- a/libs/sqf/tests/inspector.rs
+++ b/libs/sqf/tests/inspector.rs
@@ -63,9 +63,7 @@ mod tests {
     #[test]
     #[ignore = "more of a test of the wiki than of hemtt, may break on bad edits to the wiki"]
     pub fn test_wiki_examples() {
-        let file = "test_wiki_examples.sqf";
-        let mut all_examples = String::new();
-        all_examples.push_str("a=1; b=2;"); // gvars get defined in some examples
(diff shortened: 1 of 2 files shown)


libs/sqf/tests/inspector.rs
```diff
@@ -53,4 +54,83 @@ mod tests {
...
+            &hemtt_sqf::parser::database::Database::a3(false),
+            &processed,
+        )
+        .expect("for test");
+        let invalid_args = statements
+            .issues()
+            .iter()
+            .filter(|i| matches!(i, Issue::InvalidArgs(..))) // ignore unused/undefined...
+            .collect::<Vec<_>>();
+        source.vfs().remove_file().expect("vfs error");
```
PabstMirror: not sure if I'm doing something wrong but it leave behind a 
`"tests\inspector\.whiteout\test_wiki_examples.sqf_wo"` file on windows

PabstMirror: so I just gitignore it

BrettMayson: ~~This isn't a HEMTT thing... I don't know what .whiteout is~~

It is from the vfs library though, do these tests need a physical layer if the files are deleted after anyway? Can we just keep it in memory?

PabstMirror: thanks for the help on this

PR #1115, opened 2025-11-04
workspace: Handle prefix and mainprefix in `check_is_missing_file`

PabstMirror:
```
mainprefix = "z"
prefix = "mymod"
```
was getting missing files when using `z\ace\...` because they had the same mainprefix

### DIFF
--- a/libs/workspace/src/missing.rs
+++ b/libs/workspace/src/missing.rs
@@ -22,8 +22,8 @@ pub fn check_is_missing_file(target: &str, project: &ProjectConfig, processed: &
         .expect("no sources");
     let expected_path = project.expected_path();
     let target_lower = target.to_ascii_lowercase();
-    if !target_lower.starts_with(expected_path)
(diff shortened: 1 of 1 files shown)


libs/common/src/config/project/mod.rs
```diff
@@ -128,6 +131,12 @@ impl ProjectConfig {
         &self.runtime
     }
 
+    #[must_use]
+    /// Expected combined-path prefix for the project (in lowercase, without leading backslash)
+    pub fn expected_path(&self) -> &str {
```
PabstMirror, line 136: I'm not sure if you want this clutter here

it seems like it would be a tiny bit faster to just generate it once, 
but it might be cleaner to just do it in `check_is_missing_file`

BrettMayson: There might be other places in HEMTT where this can be used? Not sure off the top of my head

PR #1117, opened 2025-11-07
sqf: Inspector - check for nil/assignment types and call returns

PabstMirror:
Properly gets return values from `call`
Tracks "Assignment" as a type (I think in-game it's a GIAS error)
Using a non explicit nil will be an error
```
hashValue nil;  // ok
hashValue drawIcon3d [...] // error
```

Catches
- https://github.com/acemod/ACE3/pull/11131
```
warning[L-S12]: Invalid argument (nil) for `[U:case]`
   ┌─ addons/wardrobe/dev/compareContainerMaxLoad.sqf:22:14
   │
22 │         case default { true };
   │              ^^^^^^^ expected non-nil value
   │
   = note: found type was Nothing
```
- https://github.com/acemod/ACE3/pull/11134
```
warning[L-S12]: Invalid argument type for `[B:select]`
   ┌─ addons/medical_damage/functions/fnc_determineIfFatal.sqf:55:11
   │
55 │         } forEach _bodyPartDamage select [2];
   │           ^^^^^^^ expected Array, Config, String
   │
   = note: found type was Assignment
```


libs/sqf/tests/inspector.rs
```diff
@@ -97,6 +97,10 @@ mod tests {
 
         let workspace = hemtt_workspace::Workspace::builder()
             .memory()
+            // .physical(
+            //     &std::path::PathBuf::from(crate::ROOT),
+            //     hemtt_workspace::LayerType::Source,
+            // )
```
BrettMayson, line 103: ?

PabstMirror: just for testing, sometimes it is still nice to get it in a file
e.g. figuring out which wiki command has a syntax error

PR #1131, opened 2025-11-20
sqf: reimplementing command

BrettMayson:
image

### DIFF
--- a/libs/sqf/tests/lints.rs
+++ b/libs/sqf/tests/lints.rs
@@ -109,12 +109,14 @@ fn lint(file: &str, ignore_inspector: bool) -> (String, SqfReport) {
                 database.clone(),
             );
             (
-                codes
-                    .iter()
-                    .map(|e| e.diagnostic().unwrap().to_string(&workspace_files))
-                    .collect::<Vec<_>>()
-                    .join("\n")
-                    .replace('\r', ""),
+                {
+                    let mut codes = codes
+                        .iter()
+                        .map(|e| e.diagnostic().unwrap().to_string(&workspace_files))
+                        .collect::<Vec<_>>();
+                    codes.sort();
+                    codes.join("\n").replace('\r', "")
+                },
                 report.expect("exist"),
             )
         }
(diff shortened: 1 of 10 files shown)


libs/sqf/src/analyze/lints/s33_reimplementing_command/distance.rs
```diff
@@ -0,0 +1,235 @@
...
+        self.severity
+    }
+
+    fn message(&self) -> String {
+        if self.is_sqr {
+            String::from("code can be replaced with `distanceSqr`")
+        } else if self.dimension == 2 {
+            String::from("code can be replaced with `distance2D`")
+        } else {
+            String::from("code can be replaced with `distance`")
```
PabstMirror:
```suggestion
            String::from("code can be replaced with `vectorDistance`")
```
`distance` takes AGL and will be slightly different on each terrain

libs/sqf/src/analyze/lints/s33_reimplementing_command/atan2.rs
```diff
@@ -0,0 +1,99 @@
...
+    fn label_message(&self) -> String {
+        String::from("use `atan2`")
+    }
+
+    fn diagnostic(&self) -> Option<Diagnostic> {
+        self.diagnostic.clone()
+    }
+
+    fn suggestion(&self) -> Option<String> {
+        Some(format!("{} atan2 {}", self.left, self.right))
```
PabstMirror, line 78: atan (-5/-2) isn't the same as -5 atan2 -2
but I can't think of any reason someone would want that behavior

PabstMirror: maybe not, I knew that looked familiar 
see https://github.com/acemod/ACE3/pull/6773/files#r250479159

BrettMayson: hm, so this one might need to be cut

PabstMirror:
for `L-S33-FLOOR`
realized it handles negative number differently
r = -2.2;
r - (r % 1) = -2
floor r = -3

BrettMayson:
:|

PR #1162, opened 2025-12-29
signing: reuse private keys

BrettMayson:
Uses brute-force resistant encryption with random passwords to store the private keys, `.biprivatekey`s are not allowed.

Also checks git history, gitignore, etc. for potentially leaked or about to be leaked files


bin/src/commands/keys.rs
```diff
@@ -0,0 +1,180 @@
...
+            println!("Project:   {}", ctx.config().name());
+            println!("Prefix:    {}", ctx.config().prefix());
+            println!("Git:       {git_hash}");
+            println!("Authority: {authority}");
+            println!();
+            warn!("The generated key will be usable ONLY with this project");
+            dialoguer::Confirm::new()
+                .with_prompt("Confirm")
+                .default(false)
+                .interact()?;
```
PabstMirror:
```suggestion
            if !dialoguer::Confirm::new()
                .with_prompt("Confirm")
                .default(false)
                .interact()?
            {
                return Ok(Report::new());
            }
```

PabstMirror (approved):

PR #1180, opened 2026-01-11
Detect mismatched `else`

PabstMirror:
ref #1168

rough prototype
Will warn on mismatched else sides, (could also expand this to switch blocks)

 `private _data = if (alive _this) then { 1 } else { "They're dead" };` // warn

but will only report when the value is actually used
```
if (x) then {
    y pushBack 5
} else {
    z set ["key", value] // technically different types but not used, so no problem
};
```

but I'm seeing just a few valid uses that trigger this
ace has
```

        // check if replacement is for item type or class name
        private _configName = configName _x;
        private _toReplace = if ((_configName select [0,9]) == "ItemType_") then {
            parseNumber (_configName select [9])
        } else {
            _configName
        };
```
```
        // screenLocation [ARRAY or STRING] - Optional (will be converted to ARRAY here!)
        private _screenLocation = if (isArray (_x >> "screenLocation")) then {
            getArray (_x >> "screenLocation") // modelOffset
        } else {
            getText (_x >> "screenLocation") // memPoint
        };
```

cba has
```
#define RETDEF(VARIABLE,DEFAULT_VALUE) (if (isNil {VARIABLE}) then [{DEFAULT_VALUE}, {VARIABLE}])
#define RETNIL(VARIABLE) RETDEF(VARIABLE,nil)
(if (isNil {missionVersion}) then [{nil}, {missionVersion}])
```


PabstMirror:
moved into #1136

PR #1186, opened 2026-01-14
sqf: Add lint for `count` being replaceable with `findIf`

PabstMirror:


libs/sqf/src/analyze/lints/s35_count_skipable.rs
```diff
@@ -0,0 +1,152 @@
...
+
+Check [the wiki](https://community.bistudio.com/wiki/findIf) to learn more about `findIf` optimization.
+"
+    }
+
+    fn default_config(&self) -> LintConfig {
+        // Disabled by default because `count` could be appropriate in some cases
+        // e.g. { x setVar y; alive x} count allUnits == 0;
+        LintConfig::help().with_enabled(LintEnabled::Pedantic)
+    }
```
BrettMayson, line 47: Would be nice if the wiki could have which commands have effects or not, so HEMTT could detect this

BrettMayson (approved):

PR #1188, opened 2026-01-15
Fix error-on-all for hidden warnings from includes

PabstMirror:


DartRuffian:
Log file from me running it on a project experiencing issues.
[latest.log](https://github.com/user-attachments/files/24652059/latest.log)

> `2026-01-15T19:58:01.370319Z TRACE ThreadId(01) failed(error_on_all): codes [PW3, PW3, PW3, PW3, PW3, PW3, PW3]`

I do have a `#pragma hemtt suppress pw3_padded_arg file` in the project, so it seems like `-e` isn't accounting for stuff that's being suppressed?

PabstMirror:
If PW3 came from an /include
It would never be reported in CI or stdout but still cause the fail flag

This now only fails if things were written to stdout
(using identical `HEMTT_REPORT_WITH_INCLUDES` env logic)

PR #1190, opened 2026-01-18
Add check for pboprefixes

PabstMirror:


bin/src/modules/pboprefix.rs
```diff
@@ -0,0 +1,39 @@
...
+use crate::modules::Module;
+
+#[derive(Debug, Default)]
+pub struct PboPrefix;
+
+impl Module for PboPrefix {
+    fn name(&self) -> &'static str {
+        "pboprefix"
+    }
+    fn pre_build(
```
BrettMayson: Should probably be check stage

bin/src/modules/pboprefix.rs
```diff
@@ -0,0 +1,39 @@
...
+            {
+                return;
+            }
+            let addon_expected = format!("{}addons\\{}", project_expected, addon.name());
+            if !addon
+                .prefix()
+                .to_string()
+                .eq_ignore_ascii_case(&addon_expected)
+            {
+                warn!("Unexpected pboprefix for addon: `{}`", addon.name());
```
BrettMayson: This won't show up in CI

BrettMayson (changes requested):

Issue #1181, opened 2026-01-11
invalid `-f json` output on some config.cpp's


nobodyshome:
**Tools (complete and add to the following information):**
- **HEMTT:** `1.18.1`
- **Operating System:** Windows

**Description:**
Using `hemtt utils config convert -f json` & `hemtt utils config derapify -f json` produces invalid JSON on certain configs. Not exactly sure why, may be some nesting limit/issue?
eg:
```
                class Weather1: Weather1 {
                    sky = "A3\Map_VR\data\sky_semicloudy_sky.paa";
                    skyR = "A3\Map_VR\data\sky_semicloudy_lco.paa";
                    horizon = "A3\Map_VR\data\sky_semicloudy_horizont_sky.paa";
                };
                class Weather2: Weather1 {};
                class Weather3: Weather1 {};
                class Weather4: Weather1 {};
                class Weather5: Weather1 {};
                class Weather6: Weather1 {};
                class Weather7: Weather1 {};
            };
```
Outputs:
```
                    "Weather1": {
                        "__parent": "Weather1",
                        "sky": "A3\\Map_VR\\data\\sky_semicloudy_sky.paa",
                        "skyR": "A3\\Map_VR\\data\\sky_semicloudy_lco.paa",
                        "horizon": "A3\\Map_VR\\data\\sky_semicloudy_horizont_sky.paa"
                    },
                    "Weather2": {},
                    "__parent": "Weather1"
                },
                "Weather3": {},
                "__parent": "Weather1"
            },
            "Weather4": {},
            "__parent": "Weather1"
        },
        "Weather5": {},
        "__parent": "Weather1"
    },
    "Weather6": {},
    "__parent": "Weather1"
},
"Weather7": {},
"__parent": "Weather1"
}
}
```
Also hangs when using `-f json-pretty` I'm guessing due to the bad JSON.

**Steps to reproduce:**
Run `hemtt utils config convert -f json` on the attached `config.cpp`
Parse/verify the outputted `config.json`

**Expected behavior:**
Valid JSON to be outputted. Only happens on *some* configs, attached two that failed + their JSON outputs.

[config_examples.zip](https://github.com/user-attachments/files/24553475/config_examples.zip)

PR #1195, opened 2026-01-28
JSON: fix missing parent entry on empty forward references

nobodyshome:
**Description**
Fixes invalid JSON output when converting config files with `hemtt utils config convert -f json` and `hemtt utils config derapify -f json`. Mentioned in #1181 

**Problem**
When serializing with a parent to JSON, the serialize_map() call used properties.len() as the size hint, but then also wrote a `__parent` entry. This caused the JSON serializer to close the map prematurely, resulting in malformed JSON where `__parent` fields and subsequent classes escaped to incorrect nesting levels.

```
"Weather1": {"__parent": "Weather1", "sky": "..."},
"Weather2": {},
"__parent": "Weather1"   // <-- escaped to wrong level!
```

**Fix:**
Example input
```
class Overcast: Overcast {
    class Weather1: Weather1 { sky = "..."; };
    class Weather2: Weather1 {};
    class Weather3: Weather1 {};
    class Weather4;
};
```
Before (invalid JSON)
```
"Weather1": {"__parent": "Weather1", "sky": "..."},
"Weather2": {},
"__parent": "Weather1"
},
"Weather3": {},
"__parent": "Weather1",
"Weather4": { "Weather4": null }
```
After (valid JSON)
```
"Weather1": {"__parent": "Weather1", "sky": "..."},
"Weather2": {"__parent": "Weather1"},
"Weather3": {"__parent": "Weather1"},
"Weather4": null
```

I tested against the `config_examples.zip` mentioned in the issue: #1181 
Along with ensuring working configs continued to work as intended. 

Fair warning I'm extremely new to rust. I've managed to identify the problem (i think) and attempt a fix it based on other JSON output. However syntax, the right JSON to output and best practices are unknown to me.

### DIFF
--- a/libs/config/src/model/class.rs
+++ b/libs/config/src/model/class.rs
@@ -101,9 +101,7 @@ impl serde::Serialize for Class {
                 }
                 state.end()
             }
-            Self::External { .. } => {
-                serializer.serialize_none()
-            }
+            Self::External { .. } => serializer.serialize_none(),
         }
     }
 }


nobodyshome:
> top half looks great
> 
> but I'm not sure about the change to Externals as it could break projects expecting the old style?

Makes sense! Reverted the changes to external.

PR #1196, opened 2026-01-31
sqf: Optimize `toString` for code

PabstMirror:
Allow optimizing away `toString {code}` to just the processed code string
e.g. avoid 
`configProperties [(configFile >> "CfgAmmo" >> _ammo), QUOTE(configName _x == QUOTE(QUOTE(ADDON))), false];`

need to verify that this is always safe.
arma will throw error on invalid syntax
`toString { a     b }`
but so will hemtt
is there anyway this could introduce errors?


libs/sqf/src/compiler/optimizer/mod.rs
```diff
@@ -496,4 +502,29 @@ impl Expression {
             _ => None,
         }
     }
+
+    /// Boilerplate for unary string operations
```
PabstMirror:
```suggestion
    /// Handle toString command
```

PabstMirror:
Discord discussion https://discord.com/channels/976165959041679380/1039189045655392389/1484276371516817409

This should be safe to do as it's doing the exact same thing the engine is.

PR #1210, opened 2026-02-11
config: new chumsky parser

BrettMayson:
image

I was hoping it'd be a bit faster out of the box, but ACE also isn't that config heavy


BrettMayson:
The amount of work for this to also be applied to SQF for a very small gain is just not worth it. HEMTT is too EOL

PR #1248, opened 2026-04-18
sqf: Update inspector tests for change to wiki

PabstMirror:
https://github.com/acemod/arma3-wiki/commit/84bab341c2343fbcc459e1ff24631d160a7c4190
https://github.com/acemod/arma3-wiki/commit/535f4f4a8015ace34d01a368522eda335a31ca98

`+` no longer catches mismatch because of wiki change,

the ideal solution might be to load a wiki with custom dummy commands?
so we can test flow and not the wiki itself
but this should get the tests passing for now


BrettMayson:
Ah, I'll see about fixing this on the wiki side

BrettMayson:
Should be fixed on the wiki side

PR #1253, opened 2026-05-02
config: support __EVAL in arrays

### LINKED ISSUE #1245: __EVAL in array element position causes CCHU error
- **HEMTT:** `1.18.3`
- **Operating System:** Windows 11 Pro

**Description:**
__EVAL in array element position causes CCHU error

**Steps to reproduce:**
```cpp
#define MY_MACRO(posX, vecX) \
    {__EVAL((posX) + (vecX) * cos 0)}, \
    {__EVAL((posX) + (vecX) * cos 45)}
```

**Expected behavior:**
__EVAL should be valid wherever a numeric literal is valid, including as an array
element. All arguments in the reproduction case are compile-time constants, so the
expression is fully resolvable at build time.

BrettMayson:
Close #1245


PabstMirror:
don't think it matters because it's not valid config
but `points1[] = __EVAL({1,2});` will crash with
```
ERROR panicked at libs\workspace\src\reporting\processed.rs:448:31:
begin > end (169 > 143) when slicing `class CfgPatches {
```

PR #1275, opened 2026-07-17
Common - Cleanup for rust 1.97

PabstMirror:
also noting there is a 
`this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release` 
that should be easy fix when tabled updates ref https://github.com/zhiburt/tabled/issues/583


bin/src/commands/launch/mod.rs
```diff
@@ -422,58 +422,58 @@ pub fn read_profile(
...
-                        None
-                    },
-                    Some,
-                )
-            }
-        })
-        .collect::<Option<Vec<_>>>()
-    {
+            })
+            .collect::<Option<Vec<_>>>()?;
```
PabstMirror, line 476: sorry, all this change just to use the `?` op here

libs/paa/tests/convert.rs
```diff
@@ -8,7 +8,7 @@ fn baer_to_paa() {
     assert_eq!(paa_image.maps().len(), 6);
     let mipmap = &paa_image.maps()[0].0;
     assert_eq!(mipmap.width(), 128);
-    assert!(!mipmap.is_compressed());
+    assert!(mipmap.is_compressed()); // lager than COMPRESS_THRESHOLD
```
PabstMirror, line 11: this changed with https://github.com/BrettMayson/HEMTT/commit/39c035f9355c4effcf827a1e12558e1659ec42eb

I assume it's ok

PR #1283, opened 2026-07-31
sqf: S38-S44, new lints

BrettMayson:
* **S38: DirectPrivate** Warns when a private variable is declared and then assigned in a separate statement, suggesting the use of direct initialization instead.
* **S39: ReplacedFunctions** Detects usage of BIS functions that have been replaced by native SQF commands, recommending the modern command equivalents.
* **S41: ForeachApply** Suggests replacing `forEach` loops with `apply` when the loop index variable (`_forEachIndex`) is not used.
* **S42: ForRange** Detects `for [{},{},{}]` loop patterns and recommends the more efficient `for .. from .. to .. do` syntax.

### DIFF
--- a/libs/sqf/src/analyze/lints/s44_string_concat_format.rs
+++ b/libs/sqf/src/analyze/lints/s44_string_concat_format.rs
@@ -1,0 +1,236 @@
+use std::{ops::Range, sync::Arc};
+
+use hemtt_common::config::LintConfig;
+use hemtt_workspace::{
+    lint::{AnyLintRunner, Lint, LintRunner},
+    reporting::{Code, Codes, Diagnostic, Processed, Severity},
+};
+
+use crate::{
+    BinaryCommand, Expression, Statement, UnaryCommand, analyze::LintData,
+};
+
+crate::analyze::lint!(LintS44StringConcatFormat);
+
+impl Lint<LintData> for LintS44StringConcatFormat {
+    fn ident(&self) -> &'static str {
+        "string_concat_format"
+    }
+
+    fn sort(&self) -> u32 {
+        440
+    }
+
+    fn description(&self) -> &'static str {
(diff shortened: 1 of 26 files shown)


libs/sqf/src/analyze/lints/s44_string_concat_format.rs
```diff
@@ -0,0 +1,236 @@
...
+fn collect_concat_parts(expr: &Expression, parts: &mut Vec<String>) {
+    match expr {
+        Expression::BinaryCommand(BinaryCommand::Add, lhs, rhs, _) => {
+            collect_concat_parts(lhs, parts);
+            collect_concat_parts(rhs, parts);
+        }
+        Expression::String(value, _, _) => parts.push(format!("\"{value}\"")),
+        Expression::Variable(name, _) => parts.push(name.clone()),
+        Expression::UnaryCommand(UnaryCommand::Named(cmd), inner, _) => {
+            if cmd.eq_ignore_ascii_case("str") {
```
PabstMirror: rare issue but `str` can be different with strings
ref https://github.com/CBATeam/CBA_A3/blob/master/addons/xeh/fnc_compileEventHandlers.sqf#L179

```
x = "systemChat 'bang';"; 
z = "player addEventHandler ['Fired', " + str x + "];"; 
//call compile z; // ok
z is "player addEventHandler ['Fired', ""systemChat 'bang';""];"
```

```
try: format ["player addEventHandler ['Fired', %1];", x]
z is now "player addEventHandler ['Fired', systemChat 'bang';];"
//call compile z; // error
```

if we knew the type going into `str` is not a string then it's fine
maybe just add a extra note about this to the hint

BrettMayson: For now I'll just keep the `str` around, and also disable this lint by default until we can improve it

PR #1287, opened 2026-08-02
sqf: S41 - Skip `forEach` that are assigned to a variable

PabstMirror:
skips `x = { _x + 1 } forEach [1, 2, 3];`
not perfect, will miss forEachs going directly into commands and last statement in a code block, 
but I think this should cover most of the actual times we use the return value of forEach

and the suggestion can then be suppressed by just assigning it to a var


libs/sqf/src/analyze/lints/s41_foreach_apply.rs
```diff
@@ -54,13 +53,13 @@ When a `forEach` loop never uses `_forEachIndex`, it can often be replaced with
     }
 
     fn runners(&self) -> Vec<Box<dyn AnyLintRunner<LintData>>> {
-        vec![Box::new(Runner)]
+        vec![Box::new(StatementRunner)]
```
BrettMayson: ?

PR #1296, opened 2026-08-09
sqf: S33 - Add `pushBackUnique`

LinkIsGrim:
- `if !(x in arr) then {arr pushBack x}` -> `arr pushBackUnique x`
- the same shape already using `pushBackUnique` is reported as a redundant check, since the command does it

`in` and `pushBackUnique` are both case sensitive on strings, so the rewrite is exact. Only reported when the searched array is the array pushed to, the value matches, and the branch does nothing else.


BrettMayson:
I don't think I like this or the other PRs being created as part of S33, since that's for re-implementing math that has a dedicated command, not really for these kinds of checks

PR #1302, opened 2026-08-15
sqf: Add S46-S51, command replacement lints

LinkIsGrim:
- S46 `is_equal_type`      `typeName a == typeName b` -> `a isEqualType b`
- S47 `is_equal_type_any`  chained `isEqualType` -> `isEqualTypeAny`
- S48 `is_equal_type_all`  `count`/`findIf` type checks -> `isEqualTypeAll`
- S49 `count_type`         `{_x isKindOf t} count arr` -> `t countType arr`
- S50 `count_side`         `{side _x == s} count arr` -> `s countSide arr`
- S51 `push_back_unique`   `if !(x in arr) then {arr pushBack x}` ->
                            `arr pushBackUnique x`

### DIFF
--- a/bin/src/utils/config/inspect.rs
+++ b/bin/src/utils/config/inspect.rs
@@ -44,7 +44,6 @@ pub fn inspect(file: &PathBuf) -> Result<(), Error> {
                     eprintln!("{}", diag.to_string(&workspacefiles));
                 }
             }
-            Ok(())
         }
         Err(errors) => {
             for error in errors {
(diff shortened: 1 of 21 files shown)


PabstMirror:
Don't think it's a problem, just noting that
```
{side _x == _side} count z, // 14
_side countSide z              // 13
```
the difference is countSide doesn't count groups with no units
(which seems reasonable in most cases)

LinkIsGrim:
@BrettMayson copilot review handled

Issue #1281, opened 2026-07-23
Thread main panic


Cyruz143:
**Tools (complete and add to the following information):**
- **HEMTT:** `1.20`
- **Operating System:** Fedora 44 KDE 

**Description:**
After updating HEMTT via `curl -sSf https://hemtt.dev/install.sh | sh` it's now giving this error whenever terminal is opened or any `hemtt` command is run.

```
thread 'main' (58528) panicked at hls/src/main.rs:355:10:
Failed to connect to server: Error { kind: InvalidInput, message: "invalid port value" }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

**Steps to reproduce:**

1. Using Fish, once hemtt is installed via install.sh
2. Manually fix completions in config.fish by ammending default added line too `eval (hemtt manage completions $(basename $SHELL))`
3. Witness error on terminal open

**Expected behavior:**
No errors

**Logs:**
Full stack:

```
thread 'main' (59150) panicked at hls/src/main.rs:351:40:
port is required
stack backtrace:
   0:     0x561bcb3ae6e8 - <<std[1e3c4ec04c5261a9]::sys::backtrace::BacktraceLock>::print::DisplayBacktrace as core[37f591cfbe66b0b1]::fmt::Display>::fmt
   1:     0x561bcb3c7c9a - core[37f591cfbe66b0b1]::fmt::write
   2:     0x561bcb3b3e62 - <std[1e3c4ec04c5261a9]::sys::stdio::unix::Stderr as std[1e3c4ec04c5261a9]::io::Write>::write_fmt
   3:     0x561bcb390486 - std[1e3c4ec04c5261a9]::panicking::default_hook::{closure#0}
   4:     0x561bcb3a5261 - std[1e3c4ec04c5261a9]::panicking::default_hook
   5:     0x561bcb3a54db - std[1e3c4ec04c5261a9]::panicking::panic_with_hook
   6:     0x561bcb390542 - std[1e3c4ec04c5261a9]::panicking::panic_handler::{closure#0}
   7:     0x561bcb385379 - std[1e3c4ec04c5261a9]::sys::backtrace::__rust_end_short_backtrace::<std[1e3c4ec04c5261a9]::panicking::panic_handler::{closure#0}, !>
   8:     0x561bcb39156d - __rustc[6ef78fba13ac776c]::rust_begin_unwind
[...]

PR #1314, opened 2026-08-25
hls: replace fatal panics on request paths with recoverable errors

LinkIsGrim:
Follow-up to #1312.

tower-lsp awaits notification handlers inline in the service future, so an `.expect()` anywhere they reach unwinds out of `main` and takes the whole server down.

## Fixed

| Site | Fails when |
|---|---|
| `initialized` | the client fails to return workspace folders |
| `did_create_files` / `did_delete_files` / `did_rename_files` | the client sends a URL `Url::from_str` rejects. They also parsed each URL twice; now they share a `resave` helper |
| `completion` `read_dir` | **a half-typed `#include` path** — the directory simply doesn't exist yet, so this is routine rather than exotic |
| `audio_convert` | the file has no extension, or a non-UTF-8 one |
| `paa_json` / `paa_convert` | the PAA has no maps, or its mipmap data is malformed but still passes `Paa::read` |
| `get_database` | lookup could miss between the insert and the get, or `.hemtt/commands` is unreadable or not valid UTF-8 |
| `main()` | no port, or one that can't be connected to — **#1281**, where a shell completion setup ran the binary bare and got a 21-frame backtrace on every terminal open. Now prints usage and exits 2 |

- `Database::a3_with_workspace` was documented `# Panics` and `expect`ed on reading `.hemtt/commands`. It now returns `Error::CustomCommandIo`.
- `MipMap::get_image` and `MipMap::json` were both documented `# Panics`. They are now fallible. `Paa::json` already returned `Result` and just propagates; the CLI's `convert` and `compress` report the failure rather than dying; `wasm` keeps its existing panic.

## Not changed

`.expect()`s that cannot fail are left alone.

## Testing

The `main()` behaviour is covered by `hls/tests/cli.rs`, which spawns the real binary and asserts the exit code and that stderr contains no `panicked`.

`audio::convert` is covered against the fixture `$PBOPREFIX$`.

Also verified by driving a real LSP session over the socket against the built binary. `didOpen` on a loose `.sqf`, `didRenameFiles` with no `didOpen`, and a loose `.sqf` inside a complete project all survive.


LinkIsGrim:
Introduces 8192x8192 size limit for paa file (in case the header says something stupid like 65k*65k) decoding since it broke one of the new tests (16GiB allocation on Windows -> crash). Allocation issue was already on main branch but the test caught it.

I can just drop the test though.

hls/src/positions.rs
```diff
@@ -9,8 +9,22 @@ pub trait ToPosition {
 impl ToPosition for LineCol {
     fn to_position(&self) -> Position {
         Position::new(
-            u32::try_from(self.1.0).expect("Failed to convert line to u32"),
```
BrettMayson, line 12: I'm not a fan of these ones.

There are definitely places where hemtt should not crash where it currently does, but I also prefer a crash over bad data. We should never hit this expect in normal operations, no one has files this big, and if we get bad data in that does cause this, I want a crash.

LinkIsGrim: Fixed.

get_database still won't panic but degrade to just not linting now

better than crashing IMO?

also dropped the PAA cap stuff

PR #1317, opened 2026-08-26
hls: check the same config files as the CLI

LinkIsGrim:
The CLI checks every rapifiable file in an addon - `cpp`, `rvmat`, `ext`, `sqm`, `bikb`, `bisurf` - and honors the addon's rapify settings. The language server only ever looked at `config.cpp`, because that is what `workspace.root().addons()` collects, and ignored `exclude` and `enabled` entirely.

So an error in a `.rvmat` or `.ext` was reported by `hemtt check` and never shown in the editor, unless the file happened to be `#include`d from a `config.cpp`. In the other direction, a file the project had deliberately excluded from rapifying was reported in the editor and not by the CLI.

`hemtt_config::files` now handles everything: `can_rapify` is moved out of the rapifier, and `checkable`, which walks an addon applying it along with `rapify.enabled` and the `exclude` globs. Both callers use it. The language server scans addons rather than collecting `config.cpp` paths, and skips checking with a warning if that scan fails rather than panicking.

Verified against a project built to trigger each case. `L-C08` in a `.rvmat` now appears in both; an excluded `.rvmat` appears in neither; `enabled = false` skips the addon's configs in both. On ACE3 the CLI is unchanged - 389 configs and 2746 sqf files, with the same diagnostics as before.

`hemtt-config` gains `glob`, `thiserror` and `tracing`. `glob` is not a workspace dependency yet, so it is pinned to the version `bin` already uses.


BrettMayson:
> glob is not a workspace dependency yet

If multiple parts of hemtt use something, it can become a workspace dependency

PR #1319, opened 2026-08-28
sqf: Add S53 - Detect substring comparison len mismatch

PabstMirror:


libs/sqf/src/analyze/lints/s53_select_substring_len.rs
```diff
@@ -0,0 +1,157 @@
...
+            if arr.len() != 2 {
+                return None;
+            }
+            let Expression::Number(sel_len, _) = arr[1] else {
+                return None;
+            };
+            let str_len = str.len();
+            #[allow(clippy::cast_possible_truncation, clippy::cast_sign_loss)]
+            let sel_len = sel_len.0.round() as usize;
+            println!("str_len: {str_len}, sel_len: {sel_len}");
```
PabstMirror:
```suggestion
```

PR #1333, opened 2026-09-13
pbo: Optimize O(n^2) signing to O(n)

ddiliberto123:
While signing a .pbo that contains around 180k files (90k .ogg) I noticed it takes a long time in order to sign (like 6hrs). From the profiling I did there seems to be a lot of file reads... like an excessive amount. For example, on the same 180k file pbo over 5 million reads were conducted over a 1 min span.

This has been taking a toll on my CI/CD pipeline by timing out my runners. 

I've tracked the issue down to `hash_filenames` in `read.rs`: for each header it calls self.file(), which itself scans the header list from blob_start to locate the entry. That makes the loop O(n^2). Within the same file, `gen_checksum` has the same pattern via file_raw(). The potential fix I've implemented should bring both to O(n).

I've already tested the build with my pbo. While it would previously take > 6hrs to sign my 180k file pbo, using this new build it takes < 10 min.

Tests:
```bash
Summary [   9.749s] 556 tests run: 556 passed, 1 skipped  
```

I've run the tests and linter, but I haven't verified the output signatures against the previous build. Please let me know if there are any issues.

I look forward to a reply.


ddiliberto123:
I've implemented copilot's suggestion for `hash_files`... I encapsulated the `data_offset` logic into its own helper function which `gen_checksum` and `hash_files` now use.

Tests pass locally and I also checked if the `.bisign` produced is valid under the previous HEMTT release... its valid, but unfortunately I don't have a `.pbo` that has compressed entries.

BrettMayson (approved):

BrettMayson:
This is a good catch, and a great contribution, thanks!

1mshy:
Find O(log n) and you will get into google

PR #1337, opened 2026-09-15
sqf: eval and lint config sqf strings

PabstMirror:
Idea was to gather config strings that are code and then eval and lint them.
It works but I'm not really seeing any value in it as most config code is as simple as possible.
Let me know if there is any desire to finish otherwise this can prob be closed

image


BrettMayson:
I think it is worth adding, especially if you're adding things to check for unused variables

PR #1343, opened 2026-09-20
build: Add module to convert all png/tga on build

PabstMirror:


thisconnected:
tested on https://github.com/thisconnected/Factions/tree/test_convert

around 100MB of PNGs takes ~~107 seconds to run

thisconnected:
this has brought the convert time to around 17 seconds for 100MB which is good enough for mod with some casual textures i think

PR #1345, opened 2026-09-22
config: C19 - Add note for changed classes

PabstMirror:
No clue why these are different, but this matches vanilla
Could also close this and keep the capital Cfg for consistency.

tested with
```
] apply {
    test = configFile >> _x;
    if (isClass test) then {
        if (_x isNotEqualTo configName test) then {
            systemChat format ["mismatch: %1 vs %2", _x, configName test];
        };
    };
};
```


BrettMayson:
I intentionally capitalized these

I know it is different than vanilla, but I think vanilla is wrong here

